PR #1, opened 2022-07-07
Spec for version 0.1.0

yulric:


specs/validation.md
```diff
@@ -14,4 +14,29 @@
...
             * FK - Part that is the foreign key for another table
             * naTable - Part that does not apply to the table
-    * The PK part is mandatory
\ No newline at end of file
+    * The PK part is mandatory
+
+## Validate the data type of each column in a table
+
+* Make sure the values in each column are the right type. For example, a string has not been put into a column that should only have numbers
+* Questions
```
DougManuel, line 22: Yes. Tables will have mandatory headers, but we haven't specified those yet. 
Mandatory header is an important feature. We could embed it into the parts table. I.e. have 'headerM' for mandatory. And 'header' for optional. Or we could put the mandatory headers into the validation rules object. I vote for the `headerM` option. 

also note that different programs may wish to make additional headers mandatory. This means that we have a 'core' ODM validation schema and also there can be additional user defined schema.

specs/validation.md
```diff
@@ -14,4 +14,29 @@
...
+    * Is there a document which explains what each of these data types are?
```
DougManuel, line 23: The data types are currently defined here. https://github.com/Big-Life-Lab/PHES-ODM/blob/main/metadata_en.md?plain=1#L24-L32.

However, the approach of version 2 is to make a part for everything that needs a definition. This would mean that we need to add new parts for all data types. @mathew-thomson

specs/validation.md
```diff
@@ -14,4 +14,29 @@
...
+    * How do you get the valid categories for a category data type?
```
DougManuel, line 24: All validate categories are defined by `catSetID`. 

For example, `catSetID` = fractionCat has 5 entries (at time of writing). There are three categories with partType = category. These parts are the three valid categories: liquid, mixed, solid.

specs/validation.md
```diff
@@ -14,4 +14,29 @@
...
+    * I don't think the "See unit" and "table" data types are valid for this?
```
DougManuel, line 25: Agree. @mathew-thomson looks like an area that needs fixing?

DougManuel (approved):
We need to define what the Cereberus schema object will be for these features. Meaning, that we should manually create the validation object with the rules for measureReport headers. Then @zargot's task is to generate that validation object from the parts list.

I wonder if we should create that rule set schema as a YAML object rather than a python dictionary. Not sure.

specs/validation.md
DougManuel: https://odm.discourse.group/t/mandatory-fields-for-data-reporting/38/2

specs/validation.md
DougManuel: Clarification on data types is here. https://odm.discourse.group/t/dictionary-datatypes/37

specs/validation.md
yulric: How can I find the data type for a unit, for example, the `absHumidUnits` unit?

DougManuel: In the `sets` table search, `absHumidUnits`. There is currently only one unit for this measure, `gm3`.

yulric: I was looking for a more traditional data type for each unit value, for example, `number` for `gm3`. Or is that not going to be recorded within the ODM?

specs/validation.md
```diff
@@ -0,0 +1,164 @@
+# Function Signature
+
+## Arguments
+
+The function should accept two arguments, the ODM data to be validated and the ODM data dictionary which contains the validation rules.
+
+1. The ODM data to be validated is a [dictionary](https://docs.python.org/3/tutorial/datastructures.html#dictionaries) whose keys are the table names and values are the table rows represented as a [list](https://developers.google.com/edu/python/lists) of dictionaries. For each dictionary within the list, the keys are the column names and values are the column values. For example, the data argument for a dataset consisting of the Address and Contact table can be seen below,
```
DougManuel, line 7: Just to be clear or explicit: the ODM data is most often in a CSV format, but it could also be Excel or SQL. I expect the most common workflow in python will be to import ODM data as a Pandas data frame.

We'd then validate the Pandas data frame by converting it to a dictionary? correct? I suppose that a workflow could include converting a CSV to a dictionary, etc.

specs/validation.md
```diff
@@ -0,0 +1,164 @@
...
+            "AddressTableRequired": "NA",
+            "ContactTable": "PK",
+            "ContactTableRequired": "mandatory"
+        }
+    ]
+}
+```
+and the user wants to validate the data below,
+
+```{python}
```
DougManuel, line 116: I don't see the table name in this dictionary. How do you know that this table is

specs/validation.md
```diff
@@ -0,0 +1,164 @@
...
+* **rowIndex**: The index of the error row in the table
+* **row** The dictionary containing the errorred row
+* **validationRules**: The dictionary containing the metadata fields used to produce the error
+
+For example, for the above example the function would return the following list,
+
+```{python}
+[
+    {
+        "errorType": "MissingRequiredColumn",
```
DougManuel, line 150: Why is this error "MissingRequiredColumn"? Is this an existing error within the Cerebus package or is this an ODM-specified error? If it is an ODM error, where is the error errorType defined? In which object? 

Below is an actual example of an error from the Cerebus package when there is a missing address field. I think it is totally fine to use the Cereberus default error message, but I expect, we'll need a way (a rule object) that allows us to define error and warning messages and associate them a rule or type of rule (i.e. missing table header). 


```>>> v = Validator()
>>> v.require_all
False

>>> schema = {
...   'name': {'type': 'string'},
...   'a_dict': {
...     'type': 'dict',
...     'require_all': True,
...     'schema': {
...       'address': {'type': 'string'}
...     }
...   }
... }

>>> v.validate({'name': 'foo', 'a_dict': {}}, schema)
False
>>> v.errors
{'a_dict': [{'address': ['required field']}]}

>>> v.validate({'a_dict': {'address': 'foobar'}}, schema)
True
```

DougManuel (changes requested):
1) Comment on line 150 is the only comment requiring a response. 

However, related to that comment, I think that we need a validation object that contains a description of validation rules and any additional metadata that is required for the validation. I know I've gone back and forth on this topic -- and may continue to go back and forth. The validation metadata could be a simpler version of the current validate.csv. 

- *item*: 1
- *description*: "Validate the column names in a table."
- *messageType*: "error"
- *messageText*: "MissingRequiredColumn"
- *notes*: "This rule is used to validate that the rows in a table have all their mandatory columns. For each table in the data argument, the metadata for this rule is present in the parts sheet in the partID, <table_name>Table, and <table_name>Required column."

I'm tempted to keep what is in the specification document in the final validation document. Meaning, that the ODM documentation would contain a good description and example each(??) validation rule.

2. Suggest removing the files from the repo that are not required for the minimal example. I worry that @zargot will get confused by the current validate.csv and specifications related to the first attempt at the specification.

specs/validation.md
yulric: Yes, I was envisioning this function to be the only validation function. We would then provide converter functions for the dataset in CSV, excel, SQL etc. into a Python dictionary.

specs/validation.md
yulric: The Cerberus error objects were missing some fields which I thought were required to better understand where the error was coming from and how to fix it. Made sense to me to create out own error objects.

DougManuel: I am referring to an object to store "MissingRequiredData". I was also wondering about making this more human-readable if it is to be read by a human.

yulric: The spec for the error object for this particular error is on line 136 to 143. I think it needs a description field though for a human readable message.

The error object will change depending on the error type, but most of the fields will remain the same. I was going to add on to it when adding the remaining features.

yulric:
@DougManuel @zargot Just a few more changes:

1. Renamed the validation.md file to validate-data.md
2. The second function argument is now a cerberus validation object after discussion with Doug. Updated the rest of the spec.
3. Added a new specification doc (generate-cerberus-schema.md) which goes over the creation of a cerberus object from the ODM data dictionary
4. Remove the old validation files

Let me know if you have any question.

DougManuel:
@yulric 

1) You were going to create the template for the RuleList.csv, right?

2) We've got slightly used of terms in the document such as `rules` versus `rule list`. I'll see if I can catch them and tidy up. (I made a bunch of those inconsistencies). 

3) 'meta' is great, but we should have a few other elements such as 'ruleID'.

specs/validate-data.md
```diff
@@ -0,0 +1,146 @@
...
+```
+
+with the following validation rules argument generated using the conversion function,
+
+```python
+{
+    "Address": {
+        "type": "dict", 
+        "schema": {
+            "addressID: {
```
zargot, line 80: missing close-quote

specs/validate-data.md
```diff
@@ -0,0 +1,146 @@
...
+                "type": "string", 
+                "required": True,
+                "meta": {
+                    "partID": "addressID",
+                    "AddressTableRequired": "mandatory"
+                }
+            },
+            "addL2: {
```
zargot, line 88: missing close-quote

specs/validate-data.md
```diff
@@ -0,0 +1,146 @@
...
+            "addL2": "12345 Lane Avenue"
+        }
+    ],
+    "Contact": [
+        {
+            "contactID": "1"
+        }
+    ]
+}
+```
```
zargot, line 71: cerberus says: `{'Address': ['must be of dict type'], 'Contact': ['must be of dict type']}`?

specs/validate-data.md
```diff
@@ -0,0 +1,146 @@
...
+1. The ODM data to be validated is a [dictionary](https://docs.python.org/3/tutorial/datastructures.html#dictionaries) whose keys are the table names and values are the table rows represented as a [list](https://developers.google.com/edu/python/lists) of dictionaries. For each dictionary within the list, the keys are the column names and values are the column values. For example, the data argument for a dataset consisting of the Address and Contact table can be seen below,
+    ```python
+    {
+        "Address": [
+            {
+                "addressID": "WastewaterSiteOttawa",
+                "addL1": "123 Laurier Avenue",
+                "addL2": "",
+                "city": "Ottawa",
+                "country": "Canada"
```
zargot, line 16: missing comma at the end

specs/validate-data.md
zargot: I get the same error with the larger data structure above (line 10)

zargot: in the "larger on above on line 10" I fixed it by removing the array and just having a single dictionary for Address/Contact, but in this smaller case there are two entries for Address, so that won't work.

zargot: I think I figured it out. Just adding a list in the schema seems to work.

zargot:
I'll write a proper review and delete my individual comments, sorry about the mess.

metadata/validation-rules-list.csv
```diff
@@ -0,0 +1,2 @@
+ruleId,description,messageText
```
DougManuel, line 1: I suggest additional fields for the validation-rule-list.csv. Most of these suggestions are not needed by Joakim and most follow what we are doing in the parts table.
**Additional description fields**-  label and notes in addition to the description. These three fields (labels, description and notes) are the approach used in the parts table.
**fields for versioning** - status (active, inactive), firstReleased (version number i.e. 1.0), lastUpdated, changes (a description of changes from the last version).
**messageType** - warning or error
-  Excel rule - I am not sure about this suggestion. For sure, ODM will have validation in Excel. In version 1 almost all a validation was done in Excel and we had a list of Excel validation rules. We could store the excel rule somewhere else, but I think it makes sense in this file.
- partType - Not sure, but we may want to indicate the part type: i.e. 'table' (MissingMandatoryColumn), 'measures', 'attributes', etc.
- dataFormat (wide, long or mixed) - this may be a feature we can add later because all validation for 1.0 will be for 'long' dataFormat.

yulric:
@DougManuel @zargot Updated the PR again

* I've added the additional metadata fields to the rules list. I wonder about adding a column called `dictionaryColumnsUsed` where we note down what columns in the template are used to generate the rule
* Updated the specifications to fix the bugs @zargot pointed out. Also added the specification for retrieving the table names from the dictionary.

specs/generate-cerberus-schema.md
```diff
@@ -67,31 +110,48 @@ The ODM dictionary fields used to generate the cerberus object for this rule are
...
+            "contactTable": "NA",
+            "contactTableRequired": "NA"
+        }
         {
             "partID": "addressID",
-            "AddressTable": "PK",
-            "AddressTableRequired": "mandatory",
-            "ContactTable": "NA",
-            "ContactTableRequired": "NA"
+            "addressTable": "PK",
```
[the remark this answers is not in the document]
DougManuel: Basically, everything "Table" is being removed and replaced with 'es'. 

See: https://odm.discourse.group/t/table-names-partids-vs-column-names-camelcase-vs-pascalcase/40/14

DougManuel (approved):
Approving the current state of the specifications.

Flagging:
- @yulric you were going to:
     -  generate specs/process for adding a new rule.
     - generate the specs for the `validationDocs.md` or whatever name and format we want to use for that artifact.

specs/validate-data.md
```diff
@@ -0,0 +1,161 @@
...
+### Error Report
+
+The error report will have the following fields
+
+* **errorType**: MissingMandatoryColumn
+* **tableName**: The name of the table the error object is for
+* **columnName** The name of the column the error object is for
+* **rowNumber**: The row number with the error
+* **row** The dictionary containing the errorred row
+* **validationRules**: The ODM data dictionary rule fields violated by this row
```
zargot, line 139: should we call this something else? I feel like this name implies that we are listing actual rules, but we are only listing fields that caused the rule violation.

zargot (changes requested):

.gitignore
```diff
@@ -0,0 +1,4 @@
+
+validationFunction/.DS_Store
```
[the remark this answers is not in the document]
DougManuel: Yes.

specs/generate-cerberus-schema.md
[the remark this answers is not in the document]
DougManuel: Hopefully, this issue has been fixed and all table that report observations uses plural in the table name. I.e. adressess, contacts, methods, measures...

specs/validate-data.md
DougManuel: Suggested name?? Of course, the hardest coding task is coming up with a good field name.

zargot: True. ;) What about "ruleFields" or "validationRuleFields"?

yulric:
@zargot I've made your changes, I'll merge this PR and the new-rule-template into main.

PR #3, opened 2022-07-26
New validation rule: InvalidCategory

yulric:
Adds a new validation rule that checks the values in a categorical column are part of its category set

# Question

* Should the validation be case-sensitive or the opposite?
* On the roadmap for validating lab IDs, it is currently a varchar dataType. Should we:
    1. Make it a categorical part
    2. Populate the `catSetID` column with for example `labCat`
    3. Allow the user to create their own custom dictionary where they add the allowable labID categories
   
# Future Possible Changes

* Rule IDs should be named to be the "invalid form". For example, `MissingMandatoryColumns` rather than `NoMissingMandatoryColumns`
* The `meta` field for each column in a table in the cerberus schema should be a list with the following shape for each dictionary,
    ```
    {
        "ruleId": string,
        "meta": List<Dict> which contains the rows in the ODM dictionary used to populate this part of the schema
    }
    ```
    This should make the code to populate the `validationRuleFields` column in the error report easier


validation-rules/InvalidCategory.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Error report
+
+The error report will have the following fields
+
+* **errorType**: InvalidCategory
+* **tableName**: The name of the table whose row has the invalid category
+* **columnName** The name of the column with the invalid category
+* **rowNumber**: The index of the table row with the error
+* **row** The dictionary containing the row
```
DougManuel, line 27: I didn't quite follow 'row'. We still need to have a good glossary of terms for the ODM.

validation-rules/InvalidCategory.md
```diff
@@ -0,0 +1,189 @@
...
+* **validationRuleFields**: The ODM data dictionary rule fields violated by this row
+* **message**: Invalid category <invalidCategoryValue> found in row <rowIndex> for column <columnName> in table <tableName>
```
DougManuel, line 29: Suggest also adding the value of the invalid category.

DougManuel (commented):
I gave a quick review in advance of our project meeting today. 

> Should the validation be case-sensitive or the opposite?

Yes. 

We should also validate the datatype, but that _may_ be best to consider as a separate rule, and there are a few issues to discuss coercion, etc. 

> On the roadmap for validating lab IDs, ...

Let's discuss `labCat` as a separate topic, following the validation of categories in the ODM. In short, I suggest that we keep don't generate a labCat, but keep and validate labID as varchar -- free text -- and have a separate type of validation rule that validates a user-generated list of entries in their custom dictionary. 

I suggest that we develop that rule (labID) next because there are a few design issues to consider that may affect our architecture. I like keeping user-defined validation high on our priority list, and we'll use the rule soon within the PCR analyzer application.

> Rule IDs should be named to be the "invalid form". For example, MissingMandatoryColumns rather than NoMissingMandatoryColumns. 

Good suggestion. I suggest that we implement it now.

> The meta field for each column in a table in the cerberus schema should be a list with the following shape for each dictionary,

👍

validation-rules/InvalidCategory.md
yulric: `row` is the Python object representing the row in the dataset that failed validation for this validation rule. I'll update the definition in the spec.

DougManuel:
Looks good.

PR #4, opened 2022-07-30
Specifications for using dictionary version to generate rules

DougManuel:
Each ODM version with have rules that are specific to that version. In other words, each version of the ODM has its validation schema. 

There was an important change to the organization of the dictionary from Version 1 to 2. Validation rules and schema can be applied to Version 1. This PR includes information to guide those rules. 

For discussion and not included in this PR. Currently, an ODM dictionary version is identified within GitHub version tags. This means that the parts table name does not change, and a parts table when viewed by itself does include metadata that describes the version. We currently do not have, for example, `parts_V2.0RC.csv`, we just have `parts.csv`.  

That said, there should be good hits about the version of the part.csv. For example, the largest value of 'lastUpdated' should equal the version of the parts table.


yulric:
@DougManuel 

1. Would each version 1 part use the same validation metadata as their version 2 equivalents? For example, `assayID` was replaced with `methodID`.
2. Ideally when a part is deprecated, the `lastUpdated` column would not be the same as the `firstReleased`. In your example for `comp3`, both columns have the same value , `1`, which tells me that it was deprecated in version ` and should not be included in it. Recommend making changing to `lastUpdated` to 1.1 or 2, so it can be included in the validation schema for v1 of the ODM.
3. Does the ODM use semver versioning, i.e. minor and major versions indicate breaking changes?

DougManuel:
1. Mostly, the rules are the same in version 1 as in version 2. However, there are differences in both the organization of the dictionary and the specific rules. As an example, we don't have mandatory fields in version1, but we do in version 2. As well, table names have been renamed with a few breaking changes. `measures` in version 2 is `Wwmeasures` in version 2. 

2. We followed LOINC for the released and updated versioning. (Or at least what we thought LOINC does). This versioning approach was introduced in Version 2 isn't finalized. So, we can modify the approach. I think we interpret being depreciated or replaced as not an update. Rather a depreciate is a depreciate. Currently, firstReleased and lastUpdated with the same version indicates that the part was never updated before being depreciated. 
3. Yes. Following semver versioning. However, remember it is for the entire dictionary. So, if any part of the dictionary is breaking it would be a `.1` update. i.e from 2.0 to 2.1.

yulric:
@DougManuel I've updated the specification to provide more details on the logic and also rearranged some items

@mathew-thomson Do you mind checking the if logic for backporting from version 2 to version 2 makes sense?

DougManuel:
@yulric The update looks good. 

A few notes:
- parts.csv has been updated to more closely follow semantic versioning. Now the versioning is `"2.0"` (text)where it was previously `2` (interger). 
- We need to have Cerebus schemas saved in the repository - not just the ability to make a schema. This means we need to have schemas that can validate all versions of the ODM dictionary. This doesn't mean that we need to have a schema for every ODM version update, but I think it does mean that we need a new schema whenever we make new parts, or breaking changes to parts. For example, we'd need a new schema if we make a change to mandatory headers or add a new category. However, we wouldn't need a new schema if we changed a part label. 

I envisioned we'd have a folder that would be something like 'assets' and then have `assets/rule-schema-v11.py`, `assets/rule-schema-v20.py`. Don't know. Not elegant. Not sure what people do in situations like this. Remember that all these schemas may also be updated. So, we could have a repo update for the PHES-ODM-validation from 1.0 to 1.1, which may update some of the rule-schema-vxx.py files. Feels complicated....

DougManuel:
Clarification on dictionary coding for depreciated parts. https://github.com/Big-Life-Lab/PHES-ODM/issues/229

yulric:
@DougManuel I wonder about having the validation schema saved in the validation repo as opposed to the ODM repo.

If every change to the ODM triggers a new validation schema, then putting it in the validation repo would require an update, without any update to the actual code.

specs/versioning.md
```diff
@@ -0,0 +1,40 @@
...
+
+Since the ODM follows semver versioning, breaking changes are indicated by an increment of the minor or major version. The validation schema will also follow the same approach for example ODM data v 1.1.x will be compatible with all validation schema version v1.1.x.
+
+## Validation Rules
+
+The validation rules CSV sheet encodes metadata for all the active and inactive rules implemented by the package. The CSV sheet does not have its own versioning, but is tied to the versioning of the validation package, changes to the CSV sheet will lead to updates in either the validation schemas and/or the package. 
+
+When a new validation rule is added, we need to decide which versions of the ODM the rule is compatible with. The [ODMVersion](../validation-rules//README.md/#adding-a-new-rule) field in the CSV file outlines which ODM versions the rule supported. 
+
+The addition of a new rule should be followed by a major version update of the validation package as well as an update to all supported validation schema files. The version of the validation schema files remains the same.
```
zargot, line 22: Did you mean "minor version update"?

zargot (commented):
Looks good!

specs/versioning.md
```diff
@@ -0,0 +1,40 @@
...
+
+The release of a new validation schema should result in an increment of the patch version of the validation package since no breaking changes were made.
+
+Since the ODM follows semver versioning, breaking changes are indicated by an increment of the minor or major version. The validation schema will also follow the same approach for example ODM data v 1.1.x will be compatible with all validation schema version v1.1.x.
+
+## Validation Rules
+
+The validation rules CSV sheet encodes metadata for all the active and inactive rules implemented by the package. The CSV sheet does not have its own versioning, but is tied to the versioning of the validation package, changes to the CSV sheet will lead to updates in either the validation schemas and/or the package.
```
DougManuel, line 18: Currently, rules.csv has `firstReleased`, `lastUpdated`, and `changes`. I suggest that we keep these, but they correspond to the validation package.

specs/versioning.md
```diff
@@ -0,0 +1,40 @@
...
+
+The release of a new validation schema should result in an increment of the patch version of the validation package since no breaking changes were made.
+
+Since the ODM follows semver versioning, breaking changes are indicated by an increment of the minor or major version. The validation schema will also follow the same approach for example ODM data v 1.1.x will be compatible with all validation schema version v1.1.x.
+
+## Validation Rules
+
+The validation rules CSV sheet encodes metadata for all the active and inactive rules implemented by the package. The CSV sheet does not have its own versioning, but is tied to the versioning of the validation package, changes to the CSV sheet will lead to updates in either the validation schemas and/or the package. 
+
+When a new validation rule is added, we need to decide which versions of the ODM the rule is compatible with. The [ODMVersion](../validation-rules//README.md/#adding-a-new-rule) field in the CSV file outlines which ODM versions the rule supported.
```
DougManuel, line 20: Yes. These may seem complicated and we need to check all rules for every update to see if they still work. In practice, it shouldn't be too much work -- or at least no more work than managing package dependencies when coding.

validation-rules/README.md
```diff
@@ -28,15 +28,35 @@ The steps to request or generate a new rule are:
...
+
    2.4. `messageType`: Whether this rule generates an error or a warning. The valid values are `error` and `warning`
+
    2.5. `messageText`: The error message to display. Placeholders can be placed by using <> for example <table_name>
+
    2.6. `status`: Whether this rule is active. Valid values are `active` and `inactive`
+
    2.7. `firstReleased`: Version of the validation module when this rule will be added. Uses semver versioning.
+
    2.8. `added`: The date when it was created
```
DougManuel, line 42: I don't think 'added' is needed.

DougManuel (commented):
Looks good, and the PR reflects our conversation. Minor comments.

yulric:
@DougManuel @zargot Updated based on your feedback. Can one of you approve?

DougManuel:
I started the review. So, @yulric or @zargot will need to approve. ;)

yulric (approved):

PR #5, opened 2022-08-03
add rules, schema generation and validation

zargot:
~~This is currently only a draft.~~ Please let me know if I'm on the right track. Thanks!

Shortcomings:
- Meta-fields are not yet implemented.
- ~~Package structure needs more work.~~
- ~~main.py should be replaced by a test.~~
- ~~Automatic ODM-sheet download is pending an updated link for the latest revision.~~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
+__pycache__
+
 # Mac Files
 .DS_Store
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,18 @@
+[project]
+name = "odm-validation"
+version = "0.1.0"
+# authors = [
+#     { name="Example Author", email="author@example.com" },
+# ]
+description = "Population Health and Envirmental Open Data Model (PHES-ODM) Validation Toolkit"
+readme = "README.md"
+license = { file="LICENSE" }
+requires-python = ">=3.9"
+classifiers = [
+    "Programming Language :: Python :: 3",
+    "Operating System :: OS Independent",
+]
+
+[project.urls]
+"Homepage" = "https://github.com/Big-Life-Lab/PHES-ODM-Validation"
+"Bug Tracker" = "https://github.com/Big-Life-Lab/PHES-ODM-Validation/issues"
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,2 @@
+cerberus == 1.3.4
+xlsx2csv == 0.7.8
--- a/src/odm_validation/part_tables.py
+++ b/src/odm_validation/part_tables.py
@@ -1,0 +1,132 @@
+"""Part-table definitions."""
+
+from dataclasses import dataclass
+from typing import Dict, List
+
+
+Row = dict
+Dataset = List[Row]
+Schema = dict  # A Cerberus validation schema
+
+
+@dataclass(frozen=True)
+class PartData:
+    """
+    An immutable cache of all datasets derived from the 'parts' dataset.
+    The parts-list is stripped of empty values before generating this.
+    """
+    all_rows: Dataset  # all parts
+    attr_rows: Dataset  # is_attr
+    cat_values: Dict[str, List[str]]  # Ex: ["collection"] = ["flowPr", ...]
+    table_attr_rows: Dict[str, Dataset]  # is_attr, by table name
+    table_cat_attr_rows: Dict[str, Dataset]  # is_attr && catSetID, by tbl name
+    table_rows: Dataset  # is_table
+    tables: List[str]  # table names
+
+
+# constants
+MANDATORY = "Mandatory"
+NA = {"", "NA", "Not applicable"}
+
+
+def is_table(p):
+    return p.get("partType") == "table"
+
+
+def is_attr(p):
+    return p.get("partType") == "attribute"
+
+
+def is_cat_val(p):
+    return p.get("partType") == "category"
+
+
+def is_catset_attr(p):
+    return is_attr(p) and p.get("catSetID")
+
+
+def get_partID(p):
+    return p.get("partID")
+
+
+def get_table_attr_rows(tables, attr_rows) -> dict:
+    """Returns dict of table names and their attribute parts."""
+    result = {}
+    for t in tables:
+        result[t] = []
+    for attr_row in attr_rows:
+        for t in tables:
+            if attr_row.get(t):
+                result[t].append(attr_row)
+    return result
+
+
+def get_category_values(parts, cat):
+    return list(
+        map(get_partID,
+            filter(lambda x: is_cat_val(x) and x.get("catSetID") == cat,
+                   parts)))
+
+
+def get_catset_attributes(parts: list, table: str):
+    return [x for x in parts if x.get(table) and is_catset_attr(x)]
+
+
+def strip(parts):
+    result = []
+    for row in parts:
+        result.append({k: v for k, v in row.items() if v not in NA})
+    return result
+
+
+def gen_partdata(parts) -> PartData:
+    # `parts` are stripped before processing. This is important for performance
+    # and simplicity of implementation
+    parts = strip(parts)
+
+    table_rows = [x for x in parts if is_table(x)]
+    tables = [row["partID"] for row in table_rows]
+    attr_rows = [x for x in parts if is_attr(x)]
+    table_attr_rows = get_table_attr_rows(tables, attr_rows)
+
+    cat_values = {}
+    table_cat_attr_rows = {}
+    for table in tables:
+        catset_attr_rows = get_catset_attributes(parts, table)
+        for attr_row in catset_attr_rows:
+            cat = attr_row.get("catSetID")
+            cat_values[cat] = get_category_values(parts, cat)
+        table_cat_attr_rows[table] = catset_attr_rows
+
+    return PartData(
+        all_rows=parts,
+        attr_rows=attr_rows,
+        cat_values=cat_values,
+        table_attr_rows=table_attr_rows,
+        table_cat_attr_rows=table_cat_attr_rows,
+        table_rows=table_rows,
+        tables=tables
+    )
+
+
+def init_table_schema(name, attr_schema):
+    return {
+        name: {
+            "type": "list",  # each table has a list of rows
+            "schema": {
+                "type": "dict",  # each row is a dict
+                "schema": attr_schema,
+            },
+        }
+    }
+
+
+def init_attr_schema(attr: str, cerb_rule: tuple):
+    return {
+        attr: {
+            cerb_rule[0]: cerb_rule[1],
+            # "meta": {
+            #     "partID": attr,
+            # }
+        }
+    }
--- a/src/odm_validation/rules.py
+++ b/src/odm_validation/rules.py
@@ -1,0 +1,68 @@
+"""This file defines the Rule type along with all the rule implementations."""
+
+from dataclasses import dataclass
+from typing import Callable, Tuple
+
+import part_tables as pt
+import utils
+
+
+@dataclass(frozen=True)
+class Rule:
+    """An immutable validation rule."""
+    name: str  # Rule name
+    key: str  # The Cerberus error key identifying the Rule
+    gen_schema: Callable[pt.PartData, pt.Schema]  # Cerberus schema gen. func.
+
+
+def missing_mandatory_column():
+    cerb_rule = ("required", True)
+
+    def gen_schema(data: pt.PartData):
+        schema = {}
+        for table in data.tables:
+            for attr_row in data.table_attr_rows[table]:
+                odm_rule = (table + "Required", pt.MANDATORY)
+                if attr_row.get(odm_rule[0], "").capitalize() != odm_rule[1]:
+                    continue
+                attr = pt.get_partID(attr_row)
+                attr_schema = pt.init_attr_schema(attr, cerb_rule)
+                table_schema = pt.init_table_schema(table, attr_schema)
+                utils.deep_update(table_schema, schema)
+        return schema
+
+    return Rule(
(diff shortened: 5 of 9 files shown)


yulric:
@zargot 

Overall thinking makes sense, organize the code around the validation rules. Additional comments below:

## Code Organization

* The Rule class should be expanded so that it has abstract methods for generating a cerberus schema and generating the error report. All validation rules should then extend this class and implement this method. This will ensure that all validation rules have the same structure.
* generate_cerberus_schema should be taking the parts sheet i.e. generate_rules should be incorporated into the generate_cerberus_schema function https://github.com/Big-Life-Lab/PHES-ODM-Validation/blob/main/specs/module-functions.md?plain=1#L88
* I imagine any users for the generate_cerberus_schema function will not be stripping all the sparse column values like you're doing. Can you update the generate_rules to accomodate this?
* Every rule shbould ideally have its own gen_rule_error function since the rule_error object returned can be different per rule
* validate_data should only take the following arguments https://github.com/Big-Life-Lab/PHES-ODM-Validation/blob/main/specs/module-functions.md
* Lets not hardcode column names but have a module that has constants where they're declared

## General Code Comments

* Use a requirments.txt file to automatically install all packages for the project
* Avoid using shorthand variables like a, t etc. For example, for t in table_names should be table_name in table_names. 
* Use types as much as possible. We should add a linting rule if possible that enforces this.
* Comments. You don't have to add comments to every line but it would be nice to see comments that describes the process/flow in a function.
* Use Union types to mean a variable can have one of two values https://docs.python.org/3/library/typing.html#typing.Union. For example, if a function returns a string or None, Union[string, None]

## Bugs

* Don't think your invalid category code is working. The is_catset_attr function should be using catSetID not catsetID

zargot:
I've rewritten my code and done many of the things you mentioned. Can you take another look?

DougManuel:
I would follow @yulric comments about coding, with my input being more about the purpose and organization of code. (This means that I may be off base on some comments).

### In general
- The rule object looks good. I wonder if this should include all or most aspects of rule specification such as the rule name, error, etc. Maybe even attributes like `cerb_rule`. I think this follows @yulric suggestion? It may make it easy for folks without a lot of python knowledge to generate python code for rules, and we would be very explicit about re-usable attributes between rules. The first part of the rule schema generates function would be defining the rule object based on the corresponding .md. The second part is the more custom implementation of the rule.

### Additions that we'll need sooner -- and could be added now
- We will need to store the Cerberus schema in the repo. (This may be a discussion along with schema versioning). 
- We will also need test data on the repo. I'd like that test data to be as .csv files. `measures.csv`, etc. 
- I wonder about saving the parts table in the repo. This would help be unambiguous about how a schema is being generated.
- A reproducible example of the rules as a Jupyter or Quatro notebook. This notebook would take a schema and test data and then show/return an actual Cerberus report for the rules. Meaning, we'd have test data that 'fails' at least once for each rule in the schema.

### Smaller comments
- I agree with @yurlic for the general coding comments. 
- It may just be me, but I'd put all the actual rules in a separate file with the reusable functions in utils.py or something like rules-utils.py. Eventually, we'll probably have dozens of specific rules. 
- I suggest that the rule function name for generating Cerberus be the same as the rule name. So, either we say that rule names should be snake_case (to align with the current function names for rules), or we say that rule names should be camelCase and then rename the rule functions to camelCase. 

### Throwing it out there
- Parsing parts. Just throwing this out there in these first rules.... Would turning `parts` into a dictionary or, better yet, an object be helpful? An object with attributes such as part.name, part.label, etc. One reason I say this is that you realize that not all headers apply to all parts.... that there is a lot of NAs. A key structure for parts with `partType` with attributes dependent on partType. ... that said, the flat table and functional approach `is_` and `get_` from the flat table works really well - probably best.

yulric:
> I've rewritten my code and done many of the things you mentioned. Can you take another look?

Yes, this looks like what I had in mind and more importantly I believe it will scale well when it comes to adding new rules/custom user rules.

I still think that the `_gen_report_entry` function should go into the `Rule` object, just like `gen_schema`. Each rule will have its own error report with some similarities between them. For example, right now, the `invalidCategoyValue` is not being added to the error report for the `InvalidCategoryRule`.

yulric:
@zargot You can go ahead and finish up the entire implementation and make a PR for it.

zargot:
It's cleaned up and ready for review now.

What should we do about the rule names tho?
- `def invalid_category():` current. We may change the rule name to this like @DougManuel mentioned.
- `def InvalidCategory():` would break convention and look like a class name.
- `def invalidCategory():` may be ok, but is again breaking convention.
- `def <prefix>_InvalidCategory():` may be a little verbose and isn't equal to the rule name anyway.

DougManuel:
> What should we do about the rule names tho?
> 
> * `def invalid_category():` current. We may change the rule name to this like @DougManuel mentioned.
> * `def InvalidCategory():` would break convention and look like a class name.
> * `def invalidCategory():` may be ok, but is again breaking convention.
> * `def <prefix>_InvalidCategory():` may be a little verbose and isn't equal to the rule name anyway.
> 
> Edit: We may also just ignore the difference between the rule names in documentation and code, and just write a single notice about how the rules are implemented in code. For example:
> 
> > The rules are currently implemented as functions in rules.py. They are directly named after the rules, using snake_case. For example, the rule "Invalid Category" becomes `invalid_category`.

What about changing the names and conventions in [validation-rules-list](https://github.com/Big-Life-Lab/PHES-ODM-Validation/blob/main/metadata/validation-rules-list.csv) to snake case? 
MissingMandatoryColumn --> missing_mandatory_column

DougManuel:
I've added @martinwellman @mathew-thomson to the PR. Martin will likely be the first user of the library.

DougManuel:
1) @yulric do you agree changing the naming convention of the rules to snake case? If so, @zargot go ahead and make those changes to the actual ids in validation-rules-list.

DougManuel:
Would you mind adding more documentation to the classes? I was having a hard time following what the attribute names meant. I didn't have time to go through all the code to create an interpretation, but then I thought that I shouldn't have to go through the core to interpret! 

I like the approach below from [real python](https://realpython.com/documenting-python-code/#commenting-code-via-type-hinting-python-35).

```class Animal:
    """
    A class used to represent an Animal

    ...

    Attributes
    ----------
    says_str : str
        a formatted string to print out what the animal says
    name : str
        the name of the animal
    sound : str
        the sound that the animal makes
    num_legs : int
        the number of legs the animal has (default 4)

    Methods
    -------
    says(sound=None)
        Prints the animals name and what sound it makes
    """
```

DougManuel:
The unit testing looks good. Just a comment: I questioned whether we want all those tests because it feels duplicative having very similar code in both the .md files and the testing folder. But then I thought, yes, we do want both because they serve different purposes with the .md files being specification and documentation and unit testing being, er, testing.

DougManuel:
On the topic of documentation. I still think we should also add an executable example of the rule as a Quatro or Juypter notebook. This is worth a discussion. I've viewed this as not unit testing but a 'how to' documentation. 

We generally try to follow the approach of the [Divio](https://documentation.divio.com) documentation approach. I think the executable examples would go high up in the file structure or be incorporated into the individual rule md files. (covert the `.md` files to, say, a `.ipynb` or `.qmd`. Discuss?

yulric (changes requested):
* What do you think about making the Context object only having the following fields?
	* parts - The entire parts list
	* tables - The list of tables in the parts list
	* table_rows - The list of part rows that pertains to a table
	
   The other fields don't seem to apply to all the rules, for example cat_values only applies to the InvalidCategoryRule.
* I'm not sure the [approach](https://github.com/Big-Life-Lab/PHES-ODM-Validation/pull/5/files#diff-7f2a2367249434c8aa30126870c6c1d7a4365c059dd63fb516e0c5f8621af22dR46-R47) you took for adding the invalidCategoryValue is scalable to new fields
	* It look like when we add a new rule we will need to make modifications in two places, in the validate_data and _gen_report_entry function. I think the approach we want to take is to just do all of this in the a function within the rule object like you did in the gen_schema function.
	* I'm worried if a user adds a new custom rule, your approach makes it hard for them to generate a custom error report without changing the above functions.
* I think we can organize things better:
	* A module called parts which has functions like is_table, is_attr etc.
	* A module called abstract_rule which has the Rule object
	* A module called validation_rules which has the ruleset along with the missing_mandatory_column and invalid_category objects
	* The validation module with the two functions, validate_data and generate_cerberus_schema
* In your dataclass, you have comments for the class but can you add comments to each field also
* This is my fault, can we rename value back to invalidCategoryValue?

zargot:
> * What do you think about making the Context object only having the following fields?
>   
>   * parts - The entire parts list
>   * tables - The list of tables in the parts list
>   * table_rows - The list of part rows that pertains to a table
>   
>   The other fields don't seem to apply to all the rules, for example cat_values only applies to the InvalidCategoryRule.

The idea behind the Context is to only compute derived datasets once, so that they can be reused by multiple rules. This should improve performance a lot and make it easier to write rules. The name "Context" may be confusing tho. What do you think?

> * I'm not sure the [approach](https://github.com/Big-Life-Lab/PHES-ODM-Validation/pull/5/files#diff-7f2a2367249434c8aa30126870c6c1d7a4365c059dd63fb516e0c5f8621af22dR46-R47) you took for adding the invalidCategoryValue is scalable to new fields
>   
>   * It look like when we add a new rule we will need to make modifications in two places, in the validate_data and _gen_report_entry function. I think the approach we want to take is to just do all of this in the a function within the rule object like you did in the gen_schema function.
>   * I'm worried if a user adds a new custom rule, your approach makes it hard for them to generate a custom error report without changing the above functions.

I was thinking that the error report would be generic enough to only add a couple extra common fields (like "value") once in a while, as needed. Do you think it will get polluted with specific things after a while? I agree with having all the rule-specific things in the rule functions themselves, but I think it's worth it to modify the other parts if it's generic enough and shared by multiple rules.

> * I think we can organize things better:
>   
>   * A module called parts which has functions like is_table, is_attr etc.
>   * A module called abstract_rule which has the Rule object
>   * A module called validation_rules which has the ruleset along with the missing_mandatory_column and invalid_category objects
>   * The validation module with the two functions, validate_data and generate_cerberus_schema
> * In your dataclass, you have comments for the class but can you add comments to each field also
> * This is my fault, can we rename value back to invalidCategoryValue?

The helper functions `is_table`, etc. are only used in the file they're declared in to improve the code. I agree that helper-functions used across multiple files should be in a separate helper-file, but I don't that's the case for those functions.

I have resisted using any kind of class-inheritance, methods, etc. so I don't think it makes sense to have Rule base class, unless we rewrite the code to have those things. Right now I think it's practical to have the Rule definition in the same file as the rules, for easy reference.

I realize now that rules.py is messy and that you want to split it up to make it easier to see just the actual rules, same with validation.py I suppose. That's a good point. I was attempting to use regions, but that's not really cross-editor friendly. I'll look into splitting it up like you say :+1:.

I'll also write some more comments, but I would prefer to keep that generic "value" in the error report, haha.

zargot:
> On the topic of documentation. I still think we should also add an executable example of the rule as a Quatro or Juypter notebook. This is worth a discussion. I've viewed this as not unit testing but a 'how to' documentation.

It's a good idea. What about doing it in the next PR?

DougManuel:
> > On the topic of documentation. I still think we should also add an executable example of the rule as a Quatro or Juypter notebook. This is worth a discussion. I've viewed this as not unit testing but a 'how to' documentation.
> 
> 
> 
> It's a good idea. What about doing it in the next PR?
> 
> 

Good idea. I have a bad habit of expanding PRs.

zargot:
@yulric, I've restructured the files. What do you think?

DougManuel:
Looks good to me but, again, I differ to Yulric. 

Even with the new documentation, I didn't follow the context object. However, I finally went through the code, and it now makes sense to me. `context` isn't a good name, but it is hard to create good names. 

I made modifications to the documentation for this object. Correct me if my interpretation is wrong. I didn't follow `table_catset_attr_rows`, but may be because `catsets` are somewhat awkward in the parts list. 

Context is really an object that users will not need to worry about and so I am not worried either. That said, I did wonder at one point whether it would help to turn parts into a more meaningful and organized object. The only reason why we have parts as a flat/rectangular file is because many users have a hard time with a structured object (i.e. a dictionary) AND, after much debate, we wanted to keep all partTypes in one file. I could have made more sense to split parts into the three main partTypes (measures, methods and attributes) and then organized tables separately. 

I agree that inheritance is a pain (probably not work the effort because it generates more issues than it is worth). However, you may want to consider separate objects for the main partTypes (methods, measure, attribute) and another object for tables. But don't worry, what you've got now is good and you can always refactor later. We ('er I) should at least define the properties (header) that correspond to each partType. But you've realized that some partTypes have NA for all entries for some headers (properties), because those properties are required or useful for the specific partType.

DougManuel:
I'm happy to close this PR (if @yulric is).

zargot:
@DougManuel I made some new changes now that hopefully improved the documentation and naming.

DougManuel:
That helps me quite a bit.

yulric:
@zargot I think everything looks good. Only change, Doug and I had a quick discussion and can we change `value` to `invalidValue`?

Also, regarding custom fields for the error report, yeah, right now I don't think it warrants moving the function into the Rule object but I imagine each rule will have its own custom fields in the error report. Ok, to keep it as it is for now,

DougManuel (approved):

zargot:
@yulric, I renamed 'value' now :)

zargot:
I'll squash some commits as soon as you all approve.

yulric (dismissed):

yulric (approved):

DougManuel (approved):

PR #11, opened 2022-08-19
Rename PartData fields and use constants

### LINKED ISSUE #10: Rename PartData fields and use constants
Have names for functions and `PartData` more closely adhere to the definitions in the dictionary. Instead of `is_cat_val(p)` use `is_category(p)`. Instead of `all_rows` use `all_parts`. Instead of `att_rows` use `attributes`, etc. 

_Originally posted by @DougManuel in https://github.com/Big-Life-Lab/PHES-ODM-Validation/pull/9#pullrequestreview-1076913604_

I don't mind the helper functions you have, but I agree that it would be nice if the names aligned more with the ODM. In addition, can we move all the magic strings into an enum for the partType, lots of repetitions.

_Originally posted by @yulric in https://github.com/Big-Life-Lab/PHES-ODM-Validation/issues/9#issuecomment-1219570614_

zargot:


DougManuel (approved):
Looks good to me.

src/odm_validation/part_tables.py
```diff
@@ -15,63 +16,82 @@ class PartData:
...
     catset_meta: Dict[str, list]  # meta, by catset name
-    catset_cat_rows: Dict[str, Dataset]  # is_cat, by catset name
-    catset_attr_rows: Dataset  # is_catset_attr
-    table_catset_attr_rows: Dict[str, Dataset]  # is_catset_attr, by table name
-    table_attr_rows: Dict[str, Dataset]  # is_attr, by table name
-    table_rows: Dataset  # is_table
-    tables: List[str]  # table names
+    table_catset_attr: Dict[str, Dataset]  # is_catset_attr, by tbl name
+    table_attr: Dict[str, Dataset]  # is_attr, by table name
+    table_names: List[str]
```
yulric, line 25: It looks like the table name is being duplicated between this field and the `table_attr` and `table_catset_attr` fields. What do you think about converting table_names into a dictionary and nesting its categories and attributes into it? We can then retrieve the table names from the dictionary keys.

You could also argue that the same thing is happening with the catset_values and table_catset_attr fields, where the catset names are being duplicated. 

I bring this up because we want to avoid the situation where there are two difference source of truths for the same thing, keeping them synchronized may cause bugs in the future.

yulric (changes requested):

src/odm_validation/part_tables.py
zargot: You are correct that data is duplicated. It's supposed to be a space-time trade-off, to improve performance by just accessing the information needed. I haven't profiled anything tho, and you have a good point. It's probably beneficial to have a "table" structure with table-specific data, instead of flattening it out like I've been doing. Is it ok if we merge this in as is, and I create a new PR with an attempt at restructuring?

yulric: We can always memoized or computed fields (not sure if they're a thing in Python) to save on subsequent executions. But yes, new PR works.

yulric (approved):

src/odm_validation/part_tables.py
```diff
@@ -142,16 +157,16 @@ def init_table_schema(name, attr_schema):
     }
 
 
-def init_attr_schema(rule_id: str, cerb_rule: tuple, attr_row: Row,
+def init_attr_schema(rule_id: str, cerb_rule: tuple, attr: Row,
                      meta: List[dict] = []):
-    attr = get_partID(attr_row)
+    id = get_partID(attr)
```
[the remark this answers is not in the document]
zargot: Good point. Fixed.

yulric (approved):

PR #12, opened 2022-08-22
Spec: Rename rules

### LINKED ISSUE #8: Use snake_case for rule names in the spec
> What should we do about the rule names tho?
> 
> * `def invalid_category():` current. We may change the rule name to this like @DougManuel mentioned.
> * `def InvalidCategory():` would break convention and look like a class name.
> * `def invalidCategory():` may be ok, but is again breaking convention.
> * `def <prefix>_InvalidCategory():` may be a little verbose and isn't equal to the rule name anyway.
> 
> Edit: We may also just ignore the difference between the rule names in documentation and code, and just write a single notice about how the rules are implemented in code. For example:
> 
> > The rules are currently implemented as functions in rules.py. They are directly named after the rules, using snake_case. For example, the rule "Invalid Category" becomes `invalid_category`.

What about changing the names and conventions in [validation-rules-list](https://github.com/Big-Life-Lab/PHES-ODM-Validation/blob/main/metadata/validation-rules-list.csv) to snake case? 
MissingMandatoryColumn --> missing_mandatory_column

_Originally posted by @DougManuel in https://github.com/Big-Life-Lab/PHES-ODM-Validation/issues/5#issuecomment-1212163697_

Tasks:
- Change the rule names to snake_case
- Update the invalid_category_rule error report to use invalidValue rather than invalidCategoryValue

zargot:
fixes #8


zargot:
Should the headings of `validation-rules/*.md` be in snake_case? It looks a little odd.

DougManuel (dismissed):
It isn't a biggy, but I think the rule and function names are better with the same case. Snake case, camel case or pascal case all work. 

What you've revised looks good to me. I prefer camelCase, but it seems like we aren't using that for function names.

zargot:
> In line 28 of the invalid category file, can you rename invalidCategoryValue to invalidValue?

Good catch! Fixed.

yulric (approved):

PR #14, opened 2022-08-23
add table data structure

### LINKED ISSUE #13: Duplicate table and catset structures
It looks like the table name is being duplicated between this field and the `table_attr` and `table_catset_attr` fields. What do you think about converting table_names into a dictionary and nesting its categories and attributes into it? We can then retrieve the table names from the dictionary keys.

You could also argue that the same thing is happening with the catset_values and table_catset_attr fields, where the catset names are being duplicated. 

I bring this up because we want to avoid the situation where there are two difference source of truths for the same thing, keeping them synchronized may cause bugs in the future.

_Originally posted by @yulric in https://github.com/Big-Life-Lab/PHES-ODM-Validation/pull/11#discussion_r952652139_

zargot:
Fixes #13 ~~, but keeps catset structure as is.~~

### DIFF
--- a/src/odm_validation/part_tables.py
+++ b/src/odm_validation/part_tables.py
@@ -1,7 +1,7 @@
 """Part-table definitions."""
 
 from dataclasses import dataclass
-from typing import Dict, List
+from typing import Dict, List, Set
 
 
 # type aliases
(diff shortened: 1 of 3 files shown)


zargot:
I think it will be difficult to restructure the catset data in a good way that doesn't reuse catset-keys. I would argue that it's ok to duplicate ids/keys in general, but it seems to be worth it to group the table-data.

yulric (changes requested):
Looks good but I'm wondering if the `catset_attr` field can be restructured so that the Dataset field is as below,

```python
{
    attributes: Dataset,
    values: List[str]
    catset_meta: Dict[str, list]
}
```

zargot:
Thanks, I think I made it work now. :)

yulric (approved):

PR #26, opened 2022-09-16
Added duplicate_values_found rule

yulric:


DougManuel (commented):
See comment and discussions at: https://odm.discourse.group/t/duplicate-entries-and-lastedited-field/55

DougManuel (commented):
The template is complex. I suggest that you have a section before "Steps for Measure Template" that would be "At a Glance" or "Quck Start" that is THE minimal example — list the mandatory fields and an example with these fields.

yulric:
> The template is complex. I suggest that you have a section before "Steps for Measure Template" that would be "At a Glance" or "Quck Start" that is THE minimal example — list the mandatory fields and an example with these fields.

Is this message for me or Ivan?

yulric:
@DougManuel I've updated this PR based on our discussion. 

I've also renamed the rule to `duplicate_entries_found` to signify that this only applies to the primary key column and not to any other column. Maybe `duplicate_rows_found` would be better?

DougManuel:
> > The template is complex. I suggest that you have a section before "Steps for Measure Template" that would be "At a Glance" or "Quck Start" that is THE minimal example — list the mandatory fields and an example with these fields.
> 
> Is this message for me or Ivan?

Sorry! Wrong PR. Please ignore!

DougManuel:
> Maybe duplicate_rows_found would be better?

I like duplicate entries found, but both are good. Eventually, we will have long and wide versions of ODM data, which may make duplicate_ows_found, not as an appropriate term.

specs/odm-how-tos.md
```diff
@@ -0,0 +1,71 @@
+This file specifies methods for navigating the ODM data dictionary that can be referenced by different specification files in this repo.
```
DougManuel, line 1: Thanks for creating this file. We may want something like this in the main ODM repo and/or main documentation. It is very good to have here -- at least for now.

DougManuel (approved):
Approved with a few comments and suggestions.

1) I made a commitment with suggested wording edits. Feel free to ignore/revert these suggestions. 
 2) Just a note that we discussed potentially adding a warning if there are entries that are updated and a method to filter entries in the ODM sharing toolkit to have only the most recent update. We can address these issues at a later time and in other PRs.

validation-rules/duplicate_entries_found.md
```diff
@@ -0,0 +1,259 @@
+# duplicate_entries_found
+
+This rule checks that if two or more entries in a table have the same primary key value, they have been validly updated. The rational has been described [here](https://odm.discourse.group/t/duplicate-entries-and-lastedited-field/55). In brief, the ODM wants to keep a papertrail of updates made to an entry by allowing users to update entries by: 1. Not deleting the old entry 2. Adding a new row for the updated entry and 3. Updating the `lastUpdated` field for the updated entry.
```
DougManuel, line 3: The discourse link is on a developers-only thread. We need to start making threads in public forums, but this thread has a few details, such as correcting the ERD that aren't relevant to the public. I will make a public topic on the duplicate entries topic (or feel free to make an entry if I don't). As well, the documentation will eventually discuss how to update fields. We can update the links in this .MD when the documentation is complte.

PR #29, opened 2022-09-22
Added new rules for validating minimum and maximum values for a number

yulric:
* Added new rules `less_than_min_value` and `greater_than_max_value`
* Added new how to `Getting the version 1 equivalent for a part`

Currently it does  not support dates, I just need to think that through a little bit.


validation-rules/greater_than_max_value.md
```diff
@@ -0,0 +1,190 @@
+# greater_than_max_value
+
+This rule ensure that columns values are not greater than a defined maximum value. Currently, this rule only supports columns whose data type is float or integer i.e. numbers. As an example, consider the `geoLat` column in the `sites` table which has a maximum value of 90. The following dataset snippet would fail validation,
```
DougManuel, line 3: Yes, it makes sense to focus on float and integer. Remember that there are min-max lengths for text input, but that feels like a different rule.

DougManuel (commented):
Just one issue: remember values entered into an excel sheet will record all numbers as text: `1` will usually be `'1'`. This is because different measures and methods can have different data types. There could be a categorical, boolean, or text value. Excel can only define specific data types for entire columns, as is the case for most programs. 

This means, I think, we need to validate a coerced value. However, we don't want to coerce the data during data validation, I think.

yulric:
> Just one issue: remember values entered into an excel sheet will record all numbers as text: `1` will usually be `'1'`. This is because different measures and methods can have different data types. There could be a categorical, boolean, or text value. Excel can only define specific data types for entire columns, as is the case for most programs.
> 
> This means, I think, we need to validate a coerced value. However, we don't want to coerce the data during data validation, I think.

Something to think about. Personally, I think its okay to try to coerce values to their part data type but we should probably implement type validation before that.

DougManuel:
> Something to think about. Personally, I think its okay to try to coerce values to their part data type, but we should probably implement type validation before that.

I think we are on the same page. The bottom line is that a character `'1'` should pass validation as datatype integer `1`.

yulric:
@DougManuel Can I merge this?

DougManuel:
I would prefer including a example where the value is a internet or number as a text field, i.e. '5'. 

But yes, merge away if that is implicit.

DougManuel (approved):

DougManuel (changes requested):
Sorry, I realized there is another potential issue. For quite a few measures, the minValue and maxValue are 'see unit'. This means that the measure can have different units and the minValue and maxValue depend on the unit.

We have only one unit for temperature (celsius), but consider if it had two units (celsius and Fahrenheit). The minValue when the temperature unit is celsius would be `0`, but for far Fahrenheit would be `32`.

yulric:
@DougManuel 

I made the coercion piece explicit.

Regarding the units piece, had a talk with Mathew to understand how it works. It looks like the "see unit" marker is right now at least only(?) for input part types which I've avoided touching because I need to understand them better and how they fit into validation. I say we keep this PR only about min/max values for table headers and deal with inputs a little later. I don't know the implications of doing that though, happy to change my mind.

DougManuel (approved):

yulric:
> 

I guess I'm confused again haha. When a part is a header in a table, for example `geoLat` in the `sites` table, and has a `minValue` and `maxValue`, I thought that meant that all the values for that column in the data should adhere to that min and max value?

PR #31, opened 2022-09-22
Added missing_values_found rule

yulric:
Everything should be straightforward. I added a new how-to, finding the mandatory columns for a table.


DougManuel:
Hmm. We may need a conversation on this one, and I may need to correct or clarify what I've said before. I am going to add @mathew-thomson, @il43 and @jeandavidt to this PR.

The most common scenario for missing data is that the user will forget or omit to include a value for a mandatory field. This means the most common missing data will be ``. What may also be common is `NA`. 

The missingness set will likely be used during database management or by analysts during the data cleaning process when they receive data from a data creator. In this situation, someone cleaning the data will coerce `` to one of the missingness values, such as `nr` "A value could have been recorded, however, it was not."

I don't think we need a part for an empty field, but we need a validation rule. 

What this means for this rule is that we need to check there isn't a blank entry AND that the entry isn't one of the missingness values in the missingness list. In practical terms, we need to use both the Cereberus 'nullable' and also 'forbidden'.

DougManuel:
As an aside, we may want (or will probably want) to create a process to clean data. It occurred to me that this data cleaning is part of the Cereberus package in the 'coerse' function. We'd probably want this as a separate function and Cereberus schema. Not sure. 

An example of the course or data clean function would be to mutate mandatory fields that have `` to `nr`. There would be a report and mutated data would be returned.

yulric:
@DougManuel I've updated this spec based on our talk yesterday. Can you take a look?

metadata/validation-rules-list.csv
```diff
@@ -4,4 +4,4 @@ invalid_category,Validates a categorical column,Validates that the values in a c
...
 greater than_max_value,Validates that a number value is not greater than a maximum value,Validates that a number value is not greater than a maximum value,error,Value <invalid_value> in row <row_index> in column <column_name> in table <table_name> is greater than the allowable maximum value of <max_value>,active ,v1.0.0,,all,,
-missing_values_found,Validates that a mandatory column does not have null values,The ODM dictionary provides a list of parts that are represent missing. This rule ensures that mandatory columns do not contain these values.,error,Mandatory column <column_name> in table <table_name> contains missing value <invalid_value> in row index <row_index>,,v1.0.0,,all,,
+missing_values_found,Validates that a mandatory column does not have cells with empty cells,This rule ensures that mandatory columns do not contain  empty cells,warning,Mandatory column <column_name> in table <table_name> has an empty value in row <row_index>. Was this intentional?,,v1.0.0,,all,,
```
DougManuel, line 7: "does not have cells with empty cells" --> "does not have empty cells"

validation-rules/missing_values_found.md
```diff
@@ -28,52 +43,74 @@ The following dataset snippet would pass validation
     "sites": [
         {
             "siteID": "1",
-            "geoLat": 2
+            "geoLat": "2"
```
DougManuel, line 46: Yes, identifying a string is better.

validation-rules/missing_values_found.md
```diff
@@ -28,52 +43,74 @@ The following dataset snippet would pass validation
...
 
-* **errorType**: missing_values_found
+* **warningType**: missing_values_found
 * **tableName**: The name of the table containing the missing values
 * **columnName**: The name of the mandatory column containing the missing values
 * **rowNumber**: The index of the table row with the error
 * **row**: The dictionary containing the row
 * **validationRuleFields**: The ODM data dictionary rows used to generate this rule
-* **message**: Mandatory column <column_name> in table <table_name> contains missing value <invalid_value> in row index <row_index>
+* **message**: Mandatory column <column_name> in table <table_name> has an empty value in row <row_index>. Was this intentional?
```
DougManuel, line 66: Sorry, I was unclear. For a mandatory column, there should be an error if missing, including "", `NA`, and other missingness types. 

For other non-mandatory fields (later validation rules), we should give a warning for a blank field.

DougManuel (changes requested):
Sorry, I was unclear. For a mandatory column, there should be an error if missing, including "", NA, NAN, and other missingness types.

For other non-mandatory fields (later validation rules), we should give a warning for a blank field.

yulric:
@DougManuel I've updated the PR

DougManuel (approved):
Looks good.

PR #33, opened 2022-09-27
Added specifications to add support for custom labs for MECP

yulric:
This specification is for adding a new `schema_additions` argument to the `generate_validation_schema` function.


DougManuel (approved):
Looks good. No comments.

We'll want a method to generate the schema from a user-supplied CSV or YAML file. That can be done later.

PR #35, opened 2022-09-28
Added new type_validation rule

yulric:
Files should be self-explanatory. 

* I've left datetime for a later date just like the min/max validation
* Not sure about blob?
* Haven't filled in `boolean` yet, should we get the allowed `boolean` values from the `booleanUnits` part?


validation-rules/invalid_type.md
```diff
@@ -0,0 +1,526 @@
...
+## boolean
+
+## categorical
+
+This type is handled by the [invalid_category](./invalid_category.md) rule
+
+## varchar
+
+No validation is currently needed for this type since all the previously mentioned types can be coerced into a varchar.
+
```
DougManuel, line 151: Suggest adding a section when dataType is `NA`.

I think that only partID = value can be a header with dataType = NA. We'll need a separate rule...

## NA

NA dataType is not validated with this rule. see NA_type.md TODO

DougManuel (changes requested):
See comment about dataType = NA.

validation-rules/invalid_type.md
yulric: By NA, do you mean the set of missing values that the ODM data can have?

DougManuel: We currently have `NA` in the dataType field for the partID = `value`. The value field is where we enter the value for a measure. These values and corresponding dataTypes can vary depending on the measure. For temperature, the value is a real number. For the presence of a variant, the dataType is a Booleen. 

To validate the 'value' header, we'll need to check the dataType for the partID of the measure. 

We currently indicate "NA", but we could have a different more explicit entry, such as "see measure".

DougManuel: partID = `value` will be updated to have 'see measure'.

DougManuel (approved):

yulric:
@DougManuel I've updated the PR with a note about the "see measure" data type. 

I haven't finished the spec for boolean types, can you answer the question below,

`Haven't filled in boolean yet, should we get the allowed boolean values from the booleanUnits part?`

yulric:
@DougManuel Bumping this up in case you missed my question

DougManuel:
> @DougManuel I've updated the PR with a note about the "see measure" data type.
> 
> I haven't finished the spec for boolean types, can you answer the question below,
> 
> `Haven't filled in boolean yet, should we get the allowed boolean values from the booleanUnits part?`

The validation of boolean values for data type is the same approach that we'll use for phone numbers and email, I think.
All these data types are in the parts list, but there is no provision for defining the valid data type. We _could_ add the valid data types in the parts table by making a new valid field, using the max-min field, or even the LaTeX expression field, etc. 

What do you think? 

Probably the easiest approach is to define the validation in, er, the validation.md files using regEx. There are too many data types that we need special validation, and it seems a bit hacky or an overkill to use the solutions that I describe above.

All these dataTypes can be defined using regex, which is supported by Cerberus. 

We'd need to define what is a valid true-false value. Just `TRUE` and `FALSE`, which would be `^(TRUE|FALSE)$` (I think)? Or  case insensitive `^(?i:true|false)$`? See [here](https://9to5answer.com/regex-to-match-quot-true-quot-or-quot-false-quot). 

I am inclined to be specific and say only `TRUE` and `FALSE` with a warning for other cases and an error for entries like `T` and `F`. We could coerse or mutate to `TRUE` and `FALSE` at a later time when we develop that provision or package. 

Note that TRUE and FALSE can also be category responses and that validation would be taken care of using category validation. 

That all said, boolean is a bit complicated in the parts table. I will create a Discourse topic to review with the team.

yulric:
> > @DougManuel I've updated the PR with a note about the "see measure" data type.
> > I haven't finished the spec for boolean types, can you answer the question below,
> > `Haven't filled in boolean yet, should we get the allowed boolean values from the booleanUnits part?`
> 
> The validation of boolean values for data type is the same approach that we'll use for phone numbers and email, I think. All these data types are in the parts list, but there is no provision for defining the valid data type. We _could_ add the valid data types in the parts table by making a new valid field, using the max-min field, or even the LaTeX expression field, etc.
> 
> What do you think?
> 
> Probably the easiest approach is to define the validation in, er, the validation.md files using regEx. There are too many data types that we need special validation, and it seems a bit hacky or an overkill to use the solutions that I describe above.
> 
> All these dataTypes can be defined using regex, which is supported by Cerberus.
> 
> We'd need to define what is a valid true-false value. Just `TRUE` and `FALSE`, which would be `^(TRUE|FALSE)$` (I think)? Or case insensitive `^(?i:true|false)$`? See [here](https://9to5answer.com/regex-to-match-quot-true-quot-or-quot-false-quot).
> 
> I am inclined to be specific and say only `TRUE` and `FALSE` with a warning for other cases and an error for entries like `T` and `F`. We could coerse or mutate to `TRUE` and `FALSE` at a later time when we develop that provision or package.
> 
> Note that TRUE and FALSE can also be category responses and that validation would be taken care of using category validation.
> 
> That all said, boolean is a bit complicated in the parts table. I will create a Discourse topic to review with the team.

re boolean, I believe there is a boolean part which is categorical with True or False, so we should be able to do that.

re other data types, yes, we should be defining what the valid value/values are for a type, or more appropriately a machine readable definition of the type. For email and phone, I believe a regex would be needed to define the type.

DougManuel:
Having boolean data type as a category has advantages, but it also potentially changes the structure of the data model. The whole boolean section of the data model is inconsistent and needs a review. I'll post that discussion.

DougManuel:
https://odm.discourse.group/t/boolean-data-type

DougManuel:
Good news: A decision for boolean data type! 
1) We will keep it as a data type and not have 'categories' of TRUE and FALSE.
2) We felt we should allow upper and lower case, but coerse to upper case when we a develop coerse method.

Bad news: We realized we need to support languages throughout the ODM validation package. That means parts like table names and categories will need to be valid in different languages. I'll create an issue for this.

DougManuel:
https://odm.discourse.group/t/language-support-for-categories-and-data-types/75

DougManuel:
Just a quick note: we've finally sorted out the boolean data type. The Excel parts list has been updated. A final note on Discourse will be added shortly.

We'll try to get the excel sheet over to parts v2 RC-1 ASAP.

yulric:
@DougManuel I've updated the PR with support for validating the boolean data type. Can you take a look?

DougManuel (approved):
Looks good. 

I wonder if we should make the part "TRUE" and "FALSE" but that is a discussion with the dictionary team.

DougManuel (approved):
I didn't render the .qmd file, but it looks very good.

Issue #15, opened 2022-08-24
documentation generation


zargot:
Use Sphinx to generate documentation via github actions.
- Include quarto docs, converted to markdown.
- ~~Add a project/config file for Quarto, so that we can render all quarto-parts with a single Quarto command.~~
- Make sure that the readme link anchor works (after #75).

zargot:
It seems like [Doxygen](https://doxygen.nl/manual/docblocks.html#pythonblocks) has support for python, and that the github-actions marketplace has good support for it. Should I go ahead with that or did you have something else in mind, @DougManuel @yulric ?

zargot:
Something tells me [Sphinx](https://www.sphinx-doc.org/) is a more popular choice. It seems to have good support as well.

DougManuel:
I only have a bit of experience with Doxygen - which I like. I'll defer to others.

yulric:
Personally, I would go with Sphinx, it looks nicer haha. We'll need to decide what we want to include in the doc site for this package. Currently, I think we need the API docs and the how to tutorial. Ideally, whatever doc builder we use has support for including different documentation parts into one website.

zargot:
I'll give Sphinx a try

zargot:
I've updated the issue description to reflect that we're gonna write everything in Quarto and abandon Sphinx. The reason for this is to reduce complexity and make it better in general, as a mix of two doc-generation tools became messy. A new issue has been made to handle generation of the Python package API (#76).

PR #44, opened 2022-10-14
Added toolkit demo notebook for presentation

yulric:
Has the quarto notebook with the demo I'll go through during the presentation.

@zargot I'm using the versioning branch since we want to showcase version 1 schema generation. For some reason the validationRuleFields field is not appearing in the error report? I'm guessing that's not implemented yet?


zargot:
> For some reason the validationRuleFields field is not appearing in the error report? I'm guessing that's not implemented yet?

I'll look into it.

examples/presentations/validation-demo.qmd
```diff
@@ -0,0 +1,242 @@
...
+
+```{python}
+#| echo: false
+
+import sys
+
+# Import the validation package. In the future you can install it using pip
+sys.path.append("../../src")
+
+# Allows access to all the installed packages on my computer
```
DougManuel, line 21: Wouldn't it make more sense to note there is a requirements.txt file, with instructions or a pointer to how to use the file?

examples/presentations/validation-demo.qmd
```diff
@@ -0,0 +1,242 @@
...
+
+```{python}
+from odm_validation.validation import validate_data, generate_validation_schema
+```
+
+Lets start by showing how to validate an ODM dataset. We will be working with a subset of the ODM dictionary. The dictionary has a sites table which "contains information about a site; the location where an environmental sample was taken.". The table has a number of columns but we'll be working with the `geoLat` and `geoLong` columns. These columns are mandatory in the table.
+
+The validation schema below specifies the mandatory rule.
+
+```{python}
```
DougManuel, line 41: Ideally, we've moved toward presenting the schema as a YAML file. Most of our end users will be R users. Or perhaps just good practice to use YAML instead of a python dictionary?

examples/presentations/validation-demo.qmd
```diff
@@ -0,0 +1,242 @@
...
+                "geoLong": {
+                    "required": True
+                }
+            }
+        }
+    }
+}
+```
+
+The above validation schema is actually based on a Python library called [cerberus](https://docs.python-cerberus.org/en/stable/schemas.html). Cerberus does all the validation heavy lifting in the package, the package is more a wrapper around it, making sure that the inputs and outputs to the package integrate well with the ODM.
```
DougManuel, line 60: Just an editorial comment that I found it difficult to follow the sentence subject in the paragraph. It was the use of 'it' and sentence parts like "in the package". Which package? The validation or Cerberus? I've made a commit with suggestions.

examples/presentations/validation-demo.qmd
```diff
@@ -0,0 +1,242 @@
...
+        }
+    }
+}
+```
+
+The above validation schema is actually based on a Python library called [cerberus](https://docs.python-cerberus.org/en/stable/schemas.html). Cerberus does all the validation heavy lifting in the package, the package is more a wrapper around it, making sure that the inputs and outputs to the package integrate well with the ODM. 
+
+The dataset in the next chunk has a `sites` table whose first row is invalid, its missing the `geoLat` column.
+
+```{python}
```
DougManuel, line 64: Similar comment as the above schema dictionary. I think this is fine for now, but the test data as a CSV would be better. Didn't @zargot update to a CSV in his tutorial?

examples/presentations/validation-demo.qmd
```diff
@@ -0,0 +1,242 @@
...
+from odm_validation.validation import validate_data, generate_validation_schema
+```
+
+Lets start by showing how to validate an ODM dataset. We will be working with a subset of the ODM dictionary. The dictionary has a sites table which "contains information about a site; the location where an environmental sample was taken.". The table has a number of columns but we'll be working with the `geoLat` and `geoLong` columns. These columns are mandatory in the table.
+
+The validation schema below specifies the mandatory rule.
+
+```{python}
+validation_schema = {
```
DougManuel, line 42: This object gave me an error. It worked when I added "schema" before "sites" to the dictionary.

examples/presentations/validation-demo.qmd
```diff
@@ -0,0 +1,242 @@
...
+            "geoLong": 1
+        }
+    ]
+}
+```
+
+Consequentially, when we try to validate the dataset with our constructed validation schema it should not pass. Lets test that in the next chunk.
+
+```{python}
+validation_result = validate_data(
```
DougManuel, line 81: I had an error running this function.

```
--------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
/Users/dmanuel/github/PHES-ODM-Validation/examples/presentations/validation-demo.qmd in line 1
----> 44 validation_result = validate_data(
      45     validation_schema,
      46     invalid_odm_dataset
      47 )

File ~/github/PHES-ODM-Validation/examples/presentations/../../src/odm_validation/validation.py:148, in validate_data(schema, data, data_version)
    143                     for e in attr_errors:
    144                         errors.append(_gen_error_entry(e, row))
    146 return ValidationReport(
    147     data_version=data_version,
--> 148     schema_version=schema["schemaVersion"],
    149     package_version=__version__,
    150     errors=errors,
    151 )

KeyError: 'schemaVersion'
```

examples/presentations/validation-demo.qmd
```diff
@@ -0,0 +1,242 @@
...
+    validation_schema,
+    invalid_odm_dataset
+)
+```
+
+The validation function needs two pieces of information, a validation schem and the ODM dataset to validate with it. Printing the validation result below, we can see that validation failed due to the absence of a `None` value. A failed validation results in an error report, which is a machine actionable report of all the validation errors in the dataset.
+
+```{python}
```
DougManuel, line 89: I couldn't run the notebook from this point forward. But I found that @zargot approach to print for humans was nice, and you go into the metadata quickly. 

Maybe first print the message as:

```
```{python}
assert not report.valid()
for e in report.errors:
    print("- " + e["message"])
```

Then the object walks through the parts of the error message. A key message is the error message is structured so that developers can present with different formatting in different applications (web tools, lab information management systems, etc.)

DougManuel (changes requested):
I made comments in the code. 

- Note that I couldn't execute the beyond line 70. I am unsure if the error is on my part or a bug. 
- I have comments up to creating your own schema. 
      - Some of my comments apply to the whole document, such as moving to CSV and YAML instead of dictionaries. But I don't think that is a must-do suggestion right now. 
      - The bottom sections looked good, but I didn't give them as close a read as the top. I can do that once I/we sort out the execution issues.

examples/presentations/validation-demo.qmd
yulric: Sorry, this was an issue my Python was configured. Will fix it.

yulric:
@DougManuel I've pushed changes to the notebook. Ignore all the files changes and only look at the presentations folder. Changes are below:

1. Fixed the setup code
2. Use YAML and CSV files rather than Python dictionaries
3. Better pretty printing
4. Redid the wording due to second point

Let me know if you're still having issues running the code

examples/presentations/validation-demo.qmd
```diff
@@ -0,0 +1,242 @@
+---
+title: "ODM Validation Tutorial"
+format: html
+jupyter: python3
+---
+
+This document goes along with a presentation made during the weekly ODM development group meeting. It introduces and demoes the ODM validation toolkit.
```
DougManuel: demoes --> demonstrates ?

examples/presentations/validation-demo.qmd
DougManuel: Add the `rich` library to `requirements.txt`? I suggest we keep the presentation folder a permanent part of the repo (in the main branch). @yulric Were you thinking the same? If so, it makes sense to add rich to requirements.txt>

examples/presentations/validation-demo.qmd
DougManuel: Looks good!

examples/presentations/validation-demo.qmd
DougManuel: 👍

DougManuel (approved):
Looks good. Just a few optional comments for consideration. This presentation complements the existing tutorial. I would call this document an explanatory document. 

@zargot @yulric I suggest we keep this as a permanent part of the package and add this document to the Shpinx documentation. (You may want to modify the first paragraph to say this example explains how validation works.

zargot:
> @zargot @yulric I suggest we keep this as a permanent part of the package and add this document to the Shpinx documentation. (You may want to modify the first paragraph to say this example explains how validation works.

Yeah, I agree that it would be good to merge it with the existing tutorial in some way. I can create an issue for it, as it probably requires rewriting a lot of it. Maybe we can wait with completing this PR until after the presentation, and incorporate that issue then.

DougManuel:
> > @zargot @yulric I suggest we keep this as a permanent part of the package and add this document to the Shpinx documentation. (You may want to modify the first paragraph to say this example explains how validation works.
> 
> Yeah, I agree that it would be good to merge it with the existing tutorial in some way. I can create an issue for it, as it probably requires rewriting a lot of it. ~Maybe we can wait to complete this PR until after the presentation, and incorporate that issue then.~

I like both tutorials. They seem complimentary to me. It is good to review what makes a good tutorial or an explanatory document. This is my go to. We strive to have all four types of documentation. I think tutorials are good when they are short and about a specific task. So, I'd probably split out, for example, making a schema from validating data. Have that as two tutorials. However, this notebook supports a presentation, so it will cover more topics and in more detail.

zargot:
> I like both tutorials. They seem complimentary to me. It is good to review what makes a good tutorial or an explanatory document. This is my go to. We strive to have all four types of documentation. I think tutorials are good when they are short and about a specific task. So, I'd probably split out, for example, making a schema from validating data. Have that as two tutorials. However, this notebook supports a presentation, so it will cover more topics and in more detail.

If we can keep both that's great. Then it only depends on completing #15.

PR #62, opened 2022-11-10
tests: use asset files

zargot:
Replace hardcoded parts/schemas in tests with parts/schemas loaded from the assets directory.

The remaining asset files will be incorporated in #69.


yulric (changes requested):
1. I'm not sure what happened but it looks like the `pprint_csv_file` function is no longer printing to console? Is that happening for you?
2. Can you update all the `invalid_category` rule files (assets and qmd) with the changes you made to the parts sheet?

zargot:
blocked by #64.

zargot (commented):
blocked by #64

src/odm_validation/part_tables.py
```diff
@@ -134,15 +134,17 @@ def get_mappings(part: dict, version: Version) -> List[PartId]:
...
     kind = V1_KIND_MAP.get(loc)
     if kind == MapKind.TABLE:
+        table = part[V1_TABLE]
         ids = [table]
     elif kind == MapKind.ATTRIBUTE:
         ids = [part[V1_VARIABLE]]
     elif kind == MapKind.CATEGORY:
         ids = parse_version1Category(part[V1_CATEGORY])
+    else:
+        ids = [get_partID(part)]  # no mapping (default)
```
yulric, line 147: This should probably throw/return an error since its essentially saying "we don't recognize this location".

yulric (changes requested):

src/odm_validation/part_tables.py
zargot: it ends up there if the part doesn't have any version1 information, which leads to simply falling back to the original id. Don't we want that?

DougManuel:
> For each test, can you use all the datasets and error report in the assets directory? I put them in there so we can test what I hope are all the cases.

I am not sure if my comment is the same as @yulric, but I too was wondering about the tests.
- I wondered whether the test for each rule should go in the folder for the validation rule. This is a small point. I also see the value of keeping tests together.
- There was hard coded test data in the test code. Was this intentional. I was expecting you would use the invalid data in the corresponding validation-rules folder.

DougManuel:
I suggest deleting the `metadata` folder and putting the `validation-rule-list.csv` into the `validation-rules` folder. Just in the main directory, `validation-rules/validation-rule-list.csv`.

If you do want to keep the `metadata` folder, I'd rename it because really much of the repo is metadata -- so it is hard to know what to expect is in this specific folder.

DougManuel:
This may need a discussion, but hear me out. What about moving the validation .qmd files into their respective folders within assets?

We've got a small problem that we have to 'valiation-rules' folders. One in `assets` and the other in the parent directory. 

As a user, it would be nice to be able to go into rule in `assets/validtion-rules` and find a qmd that helps explain what all the assets are.

DougManuel (changes requested):
The changes look excellent. 

See my comments about the organization of the assets. 
1) It is tempting to validation rule .qmd, tests into the single asset folder -- or at least together in a folder.   

Then put the validation rule README and validation-rule-list.csv in the parent directory of `assets/validation-rules`.

2) as an aside, tests/common.py may be better placed in src and renamed something like test_utils.

3) I noted that some parts have names that have changed in the most common part table. We _may_ want update partID names when we have a version 2 completed, especially if we have the schemaVersion indicate '2.0.0' as it does currently. The validation examples should work without updating the names because they are all self-contained -- perfect. However, there are a few sections of the rules that may break, such as reference to the missingness partIDs, which have changed in the most recent parts table.

src/odm_validation/part_tables.py
zargot: added #70

zargot:
> * value of keeping tests togethe

zargot:
> * I wondered whether the test for each rule should go in the folder for the validation rule. This is a small point. I also see the value of keeping tests together.

This is a good question. The tests aren't part of the main package. I would suggest moving test assets into `/tests/assets`, but they are also used by the documentation which will go in the `/docs` folder (eventually), so I think it makes the most sense to keep it as is (`/assets`), since it's a shared resource.

> * There was hard coded test data in the test code. Was this intentional. I was expecting you would use the invalid data in the corresponding validation-rules folder.

Will fix in #69

zargot:
> I suggest deleting the `metadata` folder and putting the `validation-rule-list.csv` into the `validation-rules` folder. Just in the main directory, `validation-rules/validation-rule-list.csv`.
> 
> If you do want to keep the `metadata` folder, I'd rename it because really much of the repo is metadata -- so it is hard to know what to expect is in this specific folder.

I agree with doing something about the metadata folder. I'm about to move `/validation-rules` into `/docs`, and the rule-list csv is more of an asset than documentation, so maybe move it to `/assets`? I'll make a new issue for this discussion.

zargot:
> The changes look excellent.
> 
> See my comments about the organization of the assets.
> 
> 1. It is tempting to validation rule .qmd, tests into the single asset folder -- or at least together in a folder.
> 
> Then put the validation rule README and validation-rule-list.csv in the parent directory of `assets/validation-rules`.

I think my previous comments cover this.

> 2. as an aside, tests/common.py may be better placed in src and renamed something like test_utils.

I think `src/odm_validation` and `tests` are supposed to be independent, so I would prefer to keep test-specific code in the `tests` dir.

> 3. I noted that some parts have names that have changed in the most common part table. We _may_ want update partID names when we have a version 2 completed, especially if we have the schemaVersion indicate '2.0.0' as it does currently. The validation examples should work without updating the names because they are all self-contained -- perfect. However, there are a few sections of the rules that may break, such as reference to the missingness partIDs, which have changed in the most recent parts table.

Yeah, we should update partIDs when releasing v2 (or rc2). I've tried to keep everything consistent with the v2-rc1 names, and specified the '-rc.1' suffix in the version strings. There may still be some inconsistencies, but I'll fix them when I see them.

yulric (approved):

DougManuel (approved):

PR #64, opened 2022-11-17
Min/max value rule refactor

yulric:
Updated the min/max value rules based on discussion with Doug and the rules for generating an error and a warning. In addition,

1. Fixed all the snippets so that they're complete
2. Updated module functions spec file to include warnings in the return
3. Updated all example files
4. Updated the metadata file
     1. `messageType` value can now be warning
     2. The `messageText` column can support multiple values separated by a ";" 

@DougManuel I think you only need to review the following files,

validation-rules/less_than_min_value.qmd
validation-rules/greater_than_max_value.qmd
metadata/validation-rules-list.csv
specs/module-functions.md


validation-rules/greater_than_max_value.qmd
```diff
@@ -26,26 +26,45 @@ whereas the following dataset snippet would pass validation,
 
 ```{python}
 # Using a yaml file here since in a CSV file every value is by default a string
-pprint_yaml_file(assets_path + "/valid-dataset.yml")
+pprint_yaml_file(assets_path + "/valid-dataset-1.yml")
```
DougManuel, line 29: Why is this a yml as opposed to a csv?

assets/validation-rules/less-than-min-value/valid-dataset-2.csv
```diff
@@ -0,0 +1,2 @@
+siteID,geoLat
+1,-89
```
DougManuel, line 2: 89 is row two number, rather than a character "89".

validation-rules/greater_than_max_value.qmd
```diff
@@ -26,26 +26,45 @@ whereas the following dataset snippet would pass validation,
...
+```
 
 ## Error report
 
-The error report will have the following fields
+This rule can generate an error and/or a warning. 
+
+An error is generated in the following cases:
+
+1. The value is a number and is greater than the max value
```
DougManuel, line 63: Grammatical suggestion. Add a period to the end of the text because they are sentences.

"The value is a number and is less than the min value" --> "The value is a number and is less than the min value."

validation-rules/greater_than_max_value.qmd
```diff
@@ -26,26 +26,45 @@ whereas the following dataset snippet would pass validation,
...
+2. The value is not a number, can be coerced into one, and is greater than the max value
+3. The value is not a number and cannot be coerced into one
+
+The error report for the first two cases is shown below,
```
DougManuel, line 67: The text within sharp brackets is not displayed in the HTML-rendered text. See below. I suggest that we display it. 

Screenshot 2022-11-23 at 5 35 57 AM

validation-rules/less_than_min_value.qmd
```diff
@@ -15,38 +15,56 @@ from odm_validation.doc_utils import (
...
 whereas the following dataset snippet would pass validation,
 
 ```{python}
-# This is using a YAML file since in a CSV file everything is a string by default
+# Using a yaml file here since in a CSV file every value is by default a string
 pprint_yaml_file(assets_path + "/valid-dataset-1.yml")
 ```
 
-If the value to be validated is a string, then it should be coerced into a number and then validated. Values that cannot be coerced should be ignored and a warning logged. For example the following snippet should fail validation,
+If the value to be validated is not a number but can be coerced into one, then validation should proceed after coercion. Whether the corced value is valid or not, a warning should reported informing the user of the mismatched type. For example the following dataset would fail validation with an error and a warning,
```
DougManuel, line 32: The error logic and messages in the error section look good. However, I found line 32 confusing, "then validation should proceed after coersion". We are coersing yet and, regardless, I would take out the reference to describing coersion.

> If the value to be validated is not a number but can be coerced into one, then validation should proceed after coercion. Whether the corced value is valid or not, a warning should reported informing the user of the mismatched type. For example the following dataset would fail validation with an error and a warning,

to 

> If  the value to be validated is not a number but can be coerced into one, then a warning is given. Whether the coerced value is valid or not, a warning should reported informing the user of the mismatched type. For example the following dataset would fail validation with an error and a warning,

DougManuel (changes requested):
1) I wondered whether we should suppress printing the load of printing functions at the top of the .qmd files. In the end, I thought let's keep it.

2) Could you make the geoLat examples have 7 digits past the decimal point? We aren't validating that data form (yet), but we may will eventually. 

2) @yulric @zargot I had a few thoughts about error and warning reporting when I reviewed this PR. I am raising now in case we want to consider them in future features.
- Essentially, all the data in a long table format (our main format) will have numbers as text. This means we will generate 1000s of warnings. We may want to have a "warning and errors summary" that would give a count of warnings and errors, without out the specific information about which lines the errors occurred. 

This means we would have the opinion of generating two summary reports: a summary report; and a detailed report. We currently follow this approach in other applications (our planning tool).

- Once we have a coersion library, we will probably want to update the warnings to give specific advice on what do to when a user receives a warning that their data can be coersed to an number. This made me wonder if we should have an addition error message section for "suggestions". Meaning, we would have "error", "warning", and "suggestion". We can always put a suggestion into an error or warning message, so this isn't really required. Just raising it as a thought or discussion point.

zargot:
> 1. I wondered whether we should suppress printing the load of printing functions at the top of the .qmd files. In the end, I thought let's keep it.

I'm cleaning it up in #72 :smile:.

> 3. @yulric @zargot I had a few thoughts about error and warning reporting when I reviewed this PR. I am raising now in case we want to consider them in future features.
> 
> * Essentially, all the data in a long table format (our main format) will have numbers as text. This means we will generate 1000s of warnings. We may want to have a "warning and errors summary" that would give a count of warnings and errors, without out the specific information about which lines the errors occurred.
> 
> This means we would have the opinion of generating two summary reports: a summary report; and a detailed report. We currently follow this approach in other applications (our planning tool).
> 
> * Once we have a coersion library, we will probably want to update the warnings to give specific advice on what do to when a user receives a warning that their data can be coersed to an number. This made me wonder if we should have an addition error message section for "suggestions". Meaning, we would have "error", "warning", and "suggestion". We can always put a suggestion into an error or warning message, so this isn't really required. Just raising it as a thought or discussion point.

Suggestions are a good idea. I would put them together with the error/warning to keep it simple and close together.

I'll make an issue for suggestions and another one for the report summary. :+1:

yulric:
> 1. I wondered whether we should suppress printing the load of printing functions at the top of the .qmd files. In the end, I thought let's keep it.
> 2. Could you make the geoLat examples have 7 digits past the decimal point? We aren't validating that data form (yet), but we may will eventually.
> 3. @yulric @zargot I had a few thoughts about error and warning reporting when I reviewed this PR. I am raising now in case we want to consider them in future features.
> 
> * Essentially, all the data in a long table format (our main format) will have numbers as text. This means we will generate 1000s of warnings. We may want to have a "warning and errors summary" that would give a count of warnings and errors, without out the specific information about which lines the errors occurred.
> 
> This means we would have the opinion of generating two summary reports: a summary report; and a detailed report. We currently follow this approach in other applications (our planning tool).
> 
> * Once we have a coersion library, we will probably want to update the warnings to give specific advice on what do to when a user receives a warning that their data can be coersed to an number. This made me wonder if we should have an addition error message section for "suggestions". Meaning, we would have "error", "warning", and "suggestion". We can always put a suggestion into an error or warning message, so this isn't really required. Just raising it as a thought or discussion point.

1. Could expand on why this should be added? Makes sense to add examples for floating point columns but I don't understand the 7 digit part.
2. Warning/Error summary, sounds good. I'll draft something in a new PR.
3.  re suggestion, what about just putting it message field?

validation-rules/greater_than_max_value.qmd
yulric: Explained in this comment, "# Using a yaml file here since in a CSV file every value is by default a string"

Expanding on that, I wanted a test case that does not generate a warning and  since in a CSV file everything is imported as a string, I had to use a YAML file here.

We could update the CSV parser to automatically strings to numbers but this seemed easier for now.

assets/validation-rules/less-than-min-value/valid-dataset-2.csv
yulric: When its imported into Python, 89 remains as a string since its a CSV file.

DougManuel:
1. I _think_ 7 digits is what is needed to resolve lat and long to a street address -- and why MECP is requiring 7 digits. We need to follow-up with our geo folks to discuss. Certainly, degrees (i.e. 88 or 89 is 60 nautical miles, right?). So, it may just look odd to people to try to validate integers.
3. Yep.

DougManuel:
> Ah I see. I wasn't writing this spec with data correctness in mind but if that's something that should be done then I'll use other columns as an example for floats and ints since I don't really want to put in 7 decimal floating points haha. Let me know if that works.

It isn't an issue - I just thought it may be helpful to have a lived example.

yulric:
@DougManuel I've updated the spec based on your feedback

* Wording and rendering fixes
* Added examples for a float data type using the geoLong column as an example

DougManuel (approved):
Looks good.

validation-rules/less_than_min_value.qmd
DougManuel: @yulric  I just approved the PR but then noticed this comment. Thoughts?

yulric: I believe I've changed that to what you wanted, links below,

https://github.com/Big-Life-Lab/PHES-ODM-Validation/blob/min-max-value-rule-ref/validation-rules/greater_than_max_value.qmd#L24

and

https://github.com/Big-Life-Lab/PHES-ODM-Validation/blob/min-max-value-rule-ref/validation-rules/less_than_min_value.qmd#L24

DougManuel: Looks good.

PR #72, opened 2022-11-22
spec: refactor use of assets in rule docs

zargot:


yulric (changes requested):
1. I was looking through the quarto docs for includes and it says the included file should be named with an underscore so its ignored when rendering, do we need to do that for setup.qmd? Doc link: https://quarto.org/docs/authoring/includes.html

yulric (approved):

DougManuel (approved):
Emoji thumbs up!

PR #75, opened 2022-11-23
spec: fix readme links

zargot:
the slash after 'README.md' caused a quarto compile error


yulric (changes requested):
Sorry, I wanted the #version1 so that it links directly to that section. Can you remove the / before the # to see if that works?

zargot:
> Sorry, I wanted the #version1 so that it links directly to that section. Can you remove the / before the # to see if that works?

yes, it seems to work!

zargot:
well, the anchor doesn't work locally for me with `quarto preview`, but it may act differently on github pages with Sphinx.

yulric (approved):

PR #83, opened 2022-12-02
move validation-rules-list.csv to assets dir

### LINKED ISSUE #71: move validation-rules-list.csv
> I suggest deleting the `metadata` folder and putting the `validation-rule-list.csv` into the `validation-rules` folder. Just in the main directory, `validation-rules/validation-rule-list.csv`.
> 
> If you do want to keep the `metadata` folder, I'd rename it because really much of the repo is metadata -- so it is hard to know what to expect is in this specific folder.

I agree with doing something about the metadata folder. I'm about to move `/validation-rules` into `/docs`, and the rule-list csv is more of an asset than documentation, so maybe move it to `/assets`?

_Originally posted by @zargot in https://github.com/Big-Life-Lab/PHES-ODM-Validation/issues/62#issuecomment-1322277081_

zargot:
fixes #71


zargot:
@yulric fixed

yulric (approved):

PR #87, opened 2022-12-08
Added new invalid_email rule spec

yulric:
Changed Files:

* Added new file validation-rules/invalid_email.qmd with the spec
* All assets for the spec are in assets/validation-rules/invalid-email
* Added the metadata for the new rule in metadata/validation-rules-list.csv

TODO:

*  Update the assets/version-2-email-columns.yaml and assets/version-1-email-columns.yaml files with the hardcoded email columns


zargot (changes requested):
I propose changing the rule name 'isEmail' to lower-case so that it's consistent with the casing of the standard cerberus rules.

Some options: email, isemail, is_email

PR #92, opened 2022-12-13
map odm datatypes to cerberus schema types

zargot:
previously we didn't map odm datatypes to cerberus datatypes during schema generation. this fixes that.

ex: `dataType: varchar` --> `type: string`


zargot:
closing, as this is supposed to be part of each rule (that needs the type to be set).

PR #94, opened 2022-12-16
Add specifications for generating warnings when generating a validation schema

yulric:
# Why

When generating a validation schema for version 1 datasets, the backported values come from version 1 columns like `version1Location` in the dictionary. Any invalid values in these columns means that we can't generate an entry in the schema for that row. These feature will provide the user with warnings when this happens.

# Files Changed

* The specification file can be seen at `specs/generate-validation-schema.qmd`.
* All the assets for the specifications are at `assets/specs/generate-validation-schema/missing-version-1-fields`.
* Updated the `module-functions.md` file since the return for the generate-validation-schema` function has changed.


docs/specs/generate-validation-schema.qmd
```diff
@@ -0,0 +1,178 @@
...
+validation schema is possible due to the efforts of the PHES-ODM team in
+creating a machine-readable data dictionary for the data.
+
+In brief, the data dictionary is a CSV/Excel file that encodes all the pieces of the data, how they're related to each other, as well as validation metadata. Using this data dictionary file, the `generate_validation_schema` function can automatically generate a validation schema for different ODM data versions.
+
+## Features
+
+At its core, the `generate_validation_schema` function creates a Python dictionary that is actionable by the `generate_validate_data` function. The details regarding the dictionary creation and the fields that go into it can be seen in the [validation-rules](../validation-rules/) folder. This document will go over the function features that are agnostic to each validation rule.
+
+### Warnings
```
DougManuel, line 27: This warning specification looks good. Please tell the dictionary development team if there are parsing errors of parts.csv. We can correct those. Version 1 won't be updated (hopefully), so correcting errors should be a limited exercise.

DougManuel: I don't even think we should generate an error if there isn't a corresponding version 1 equivalent for a version 2 part. We don't have an expectation that there should be equivalency.

yulric: > I don't even think we should generate an error if there isn't a corresponding version 1 equivalent for a version 2 part. We don't have an expectation that there should be equivalency.

Even if the `firstReleased` column is 1.0? I thought that if the previous statement is true then there has to be a version 1 equivalent part.

zargot (changes requested):
should be merged to master, and a new issue should be created, as discussed.

Issue #36, opened 2022-09-30
Add new schema_additions argument to the generate_schema function


yulric:
The specifications PR https://github.com/Big-Life-Lab/PHES-ODM-Validation/pull/33/files

zargot:
what if the user wants to use an existing schema from the repo, just with additions? Maybe we should have it as a separate function that can be called on a schema after its generation?

yulric:
I'm not sure how this new argument prevents that? Can you give an example?

Your solution is definitely much more flexible and I can go either way. We can discuss with @DougManuel.

yulric:
For sure, the most flexible option is the function approach. From a user's perspective, I *think* the argument approach is better because they can just create a JSON/YAML file with addition and use it in the function.

PR #120, opened 2023-01-23
docs: ignore generated md files in specs dir

zargot:
The repo is being polluted by this when generating docs.


zargot:
this is being fixed in #36 instead.

PR #130, opened 2023-02-03
add datetime to min/max-value rules

### LINKED ISSUE #97: Update min/max-value rules with support for date types

zargot:
fixes #97

- refactor error reporting to unify coercion and normal error logic in new module 'reports'
- fix datetime parsing/conversion
- extend min/max-rule tests with datetime

### DIFF
--- a/src/odm_validation/reports.py
+++ b/src/odm_validation/reports.py
@@ -1,0 +1,165 @@
+import datetime
+from dataclasses import dataclass, field
+from enum import Enum
+from typing import Any, List, Optional, Set
+# from pprint import pprint
+
+import part_tables as pt
+from rules import COERCION_RULE_ID
+from stdext import (
+    get_len,
+    type_name,
+)
+
+
+class ErrorKind(Enum):
+    WARNING = 'warning'
+    ERROR = 'error'
+
+
+@dataclass(frozen=True)
+class ErrorCtx:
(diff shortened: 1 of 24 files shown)


src/odm_validation/reports.py
```diff
@@ -0,0 +1,167 @@
...
+
+def gen_rule_error(ctx: ErrorCtx,
+                   err_template: Optional[str] = None,
+                   kind: Optional[ErrorKind] = None,
+                   ) -> dict:
+    """
+    :param err_template: overrides rule error-template
+    :param kind: overrides ctx.kind
+    """
+    kind = kind or ctx.kind
```
yulric, line 104: ErrorCtx does not have a `kind` field?

yulric (changes requested):
1. error-report-4.json should be reverted since its being used in the documentation and its not consistent with the dataset the report is for.
2. Ok, so one issue with the changes you made to all the assets is that the documentation is using them which means it will be confusing to users.

src/odm_validation/reports.py
zargot: done

yulric (changes requested):
1. Can you add tests for generating version 1 schemas?

yulric (approved):

PR #140, opened 2023-02-22
add batching of coercion and validation

zargot:
Cerberus can't handle big datasets. This fixes that by splitting up the workload into smaller batches.

### DIFF
--- a/src/odm_validation/reports.py
+++ b/src/odm_validation/reports.py
@@ -52,6 +52,12 @@ def _fmt_list(items: list) -> str:
         return str(items[0])
 
 
+def get_row_num(row_index: int, offset: int) -> int:
+    """Returns the spreadsheet row number of `row_index` plus `offset`,
+    converted to a one-indexed number."""
+    return offset + row_index + 1
+
+
 def _fmt_value(val: Any) -> Any:
     # needed to avoid the default `datetime.datetime(year, month, day, ...)`
     if isinstance(val, datetime.date) or isinstance(val, datetime.datetime):
--- a/src/odm_validation/validation.py
+++ b/src/odm_validation/validation.py
@@ -44,6 +44,7 @@ def _gen_cerb_rule_map():
 
 
 # private constants
+_BATCH_SIZE = 100
 _KEY_RULES = _gen_cerb_rule_map()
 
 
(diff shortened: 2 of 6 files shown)


src/odm_validation/validation.py
```diff
@@ -326,16 +333,40 @@ def _validate_data_ext(schema: Schema,
         '`data` must be a dict. Remember to wrap the datasets in a dict with '
         'the table names as keys.')
 
-    coerced_data = _coerce_data(data, coercion_schema, warnings, errors)
+    def batchTableData(action, table_id, table_data):
```
yulric, line 336: snake case?

src/odm_validation/validation.py
```diff
@@ -302,10 +304,15 @@ def generate_validation_schema(parts, schema_version=pt.ODM_VERSION_STR,
                                            schema_additions)
 
 
+# OnProgress(action, table_id, processed, total)
+OnProgress = Callable[[str, str, int, int], None]
```
yulric, line 308: The action strings should be stored in variables.

src/odm_validation/reports.py
```diff
@@ -52,6 +52,12 @@ def _fmt_list(items: list) -> str:
         return str(items[0])
 
 
+def get_row_num(row_index: int, offset: int) -> int:
```
yulric, line 55: I'm not sure this function should be in here? I see this module as being responsible for constructing a ValidationReport from ErrorCtx. 

What about moving this functionality to the getBatchData function i.e. it yield's the row number?

src/odm_validation/validation.py
zargot: not sure what you mean here? This is only a type alias, so I can't really specify names on the parameters

src/odm_validation/validation.py
zargot: done

src/odm_validation/reports.py
zargot: it's there because it's used to calculate the final row number of the validation report errors. I may also handle additional logic for adjusting the row numbers later (to match excel, etc).

The only reference to "batchdata" i can find is in validation.py, but this function is also used in cerberusext.py.

src/odm_validation/part_tables.py
```diff
@@ -33,6 +33,7 @@
 Dataset = List[Row]
 PartMap = Dict[PartId, Part]
 TableId = PartId
+TableDataset = Dict[TableId, Dataset]
```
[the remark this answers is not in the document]
zargot: done

yulric (approved):

zargot (commented):
TODO:
- write test for unique-values and prove that it doesn't work as it is now.
- fix it by keeping error-state across batches

yulric (approved):

PR #145, opened 2023-02-24
add validation error summary

### LINKED ISSUE #144: tools/validate: add a makeshift summary report
- output a new file `summary.txt` with counts for all individual errors, per table
- print the total number of errors per table at the end of program execution, as well as a reminder of where to find the output files

zargot:
fixes #144

### DIFF
--- a/src/odm_validation/validation.py
+++ b/src/odm_validation/validation.py
@@ -6,7 +6,7 @@ generation and data validation.
 from collections import defaultdict
 from copy import deepcopy
 from itertools import groupby
-from typing import Any, Callable, Dict, List, Optional, Set
+from typing import Any, Callable, Dict, List, Optional, Set, Tuple
 # from pprint import pprint
 
 from cerberusext import ContextualCoercer, OdmValidator
@@ -25,6 +25,7 @@ from stdext import (
 from versions import __version__, parse_version
 
 
+RuleError = Tuple[rules.RuleId, dict]
 TableDataset = Dict[pt.TableId, pt.Dataset]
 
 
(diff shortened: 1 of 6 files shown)


src/odm_validation/validation.py
```diff
@@ -223,7 +225,12 @@ def _strip_coerce_rules(cerb_schema):
     return strip_dict_key(deepcopy(cerb_schema), 'coerce')
 
 
-def _map_cerb_errors(cerb_errors, schema, rule_whitelist, offset: int):
+def _record_error(summary: reports.ValidationSummary, table_id, rule_id):
```
yulric, line 228: Should move this function to the `ValidationSummary` class? And rename it to `_update_report`?

tools/validate_xlsx.py
```diff
@@ -41,12 +43,40 @@ def get_sheet_table_id(schema, sheet_name) -> Optional[str]:
...
+def gen_summary(summary: reports.ValidationSummary) -> Optional[str]:
+    result = '# error summary\n'
+    total = 0
+    for table_id, table_summary in summary.table_summaries.items():
+        result += f'\n## {table_id}\n'
+        table_count_max = reduce(lambda a, b: max(a, b),
+                                 table_summary.error_counts.values())
+        width = len(str(table_count_max))
+        for rule_id, count in table_summary.error_counts.items():
+            result += f'- {count:{width}d} {rule_id}\n'
```
yulric, line 75: Add a comment here about why the width is needed.

yulric (changes requested):
1. Think we should move all the logic for creating the summary report into a seperate function, in prepearation for the new create_summary_report function that we discussed before.
2. Looking at the new code that requires passing in the ruleId is making me think we should  normalize the Error object so that it should have an error type field (Warning or Error) and a ruleId field. That way you don't have to pass it in everywhere and just get it from the object.

tools/validate_xlsx.py
zargot: done

src/odm_validation/validation.py
zargot: done, i called it count_error

yulric (approved):

PR #159, opened 2023-03-01
fix sitemeasure table

### LINKED ISSUE #152: Duplicate entries are being thrown for the siteID field in the SiteMeasure table. This is in the TMU lab data.

### LINKED ISSUE #153: Invalid category errors are being thrown for `wwFlow` values in the categorical type field in the SiteMeasure table

zargot:
fixes #152, fixes #153 

- fix multi-values in v1 mappings (related to tables in this case)
- temp fix incorrect table-attr relationship in v1 (should be fixed in next minor release)


src/odm_validation/rule_primitives.py
```diff
@@ -175,16 +161,16 @@ def gen_value_schema(data: pt.PartData, ver: Version, rule_id: str,
     :gen_cerb_rules: A function returning a dict of Cerberus rules
     """
     schema = {}
-    for table in table_items2(data):
-        table_id0 = pt.get_partID(table)
-        table_schema = init_table_schema2(schema, data, table, ver)
-        for attr in attr_items2(data, table_id0, odm_key):
+    for table_id0, table_id1, table in table_items(data, ver):
+        table_meta = get_table_meta(table, ver)
+        table_schema = init_table_schema(table_id1, table_meta, {})
```
yulric, line 166: Shouldn't `init_table_schema` choose whether to pass in table_id9 or table_id1 depending on the `ver`?

src/odm_validation/rule_primitives.py
```diff
@@ -208,16 +194,25 @@ def gen_conditional_schema(data: pt.PartData, ver: Version, rule_id: str,
...
     schema = {}
     odm_key = None
-    for table in table_items2(data):
-        table_id = pt.get_partID(table)
-        table_schema = init_table_schema2(schema, data, table, ver)
-        for attr in attr_items_pred(data, table_id, pred):
+    table_attr = {}
+    for table_id0, table_id1, table in table_items(data, ver):
+        table_meta = get_table_meta(table, ver)
+        table_schema = init_table_schema(table_id1, table_meta, {})
```
yulric, line 200: Shouldn't we pass in table_id0 in case we need to generate the schema for version 2?

yulric (changes requested):
1. I'm comparing the schemas with whats in the version. Could you check why the following were removed:
    1. collection from the Sample table
    2. siteID from the Sample table
    3. fractionAnalyzed from the WWMeasure table

src/odm_validation/rule_primitives.py
zargot: table_id0 should equal table_id1 in v2, I think, since get_mappings are defaulting to the original part id.

src/odm_validation/rule_primitives.py
zargot: same as other comment :)
table_id0 = original id
table_id1 = mapped id, which may be either v1 or v2

zargot:
> 1. I'm comparing the schemas with whats in the version. Could you check why the following were removed:
>    
>    1. collection from the Sample table

That was due to the wrong casing being used in the table name in the dictionary. It's been fixed there and I've updated the parts/schemas.

>    2. siteID from the Sample table

The only siteID (in version1Variable) i can see belongs to the Site table?
The only part with version1Variable=siteID that I can see is partID=siteID, which only has version1Table=Site, not Sample? In any case, is it still wrong now with the new update?

>    3. fractionAnalyzed from the WWMeasure table

Is it still wrong?

src/odm_validation/rule_primitives.py
yulric: Oh ok, yeah the `table_items` function return is confusing. The mapped ID should always be mapped to its version 1 equivalent or None if its does  not exist, setting it to the version 2 value is confusing, since it does not make sense to map a v2 part to a v2 part.

yulric (approved):
Thanks for looking into those parts, everything looks good now. SiteID is still an issue, but it looks like an issue in the dictionary.

src/odm_validation/rule_primitives.py
zargot: well, the idea behind the mapping was to abstract away the version and only have 'original' and 'mapped' ids, which would end up as the same id if the source and destination version was the same. This way we can write more concise logic code without as many `if version is 1` branches. Isn't that good?

yulric: Yeah, but I'm wondering if its less confusing to move all of that version mapping upstream into gen_partdata for example. Then we only have to deal with versioning in that one function.

PR #163, opened 2023-03-03
change license to CC BY 4.0

zargot:
~~CC didn't have an official plain text version on their website, and only said to write the short paragraph that I've included.~~
~~https://creativecommons.org/choose/results-one?license_code=by&amp;jurisdiction=&amp;version=4.0&amp;lang=en~~

**Source: https://creativecommons.org/licenses/by/4.0/legalcode.txt**


DougManuel (approved):
There is a github template that we use in other repositories in this project. See https://github.com/Big-Life-Lab/PHES-ODM/blob/main/LICENSE

zargot:
> There is a github template that we use in other repositories in this project. See https://github.com/Big-Life-Lab/PHES-ODM/blob/main/LICENSE

Ah, thanks! I'll use it.

PR #164, opened 2023-03-06
add rule blacklist

zargot:
and put it in a "rule filter" together with the existing whitelist

### DIFF
--- a/src/odm_validation/validation.py
+++ b/src/odm_validation/validation.py
@@ -5,6 +5,7 @@ generation and data validation.
 
 from collections import defaultdict
 from copy import deepcopy
+from dataclasses import dataclass
 from itertools import groupby
 from typing import Any, Callable, Dict, List, Optional, Set, Tuple
 # from pprint import pprint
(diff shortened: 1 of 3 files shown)


src/odm_validation/validation.py
```diff
@@ -29,6 +30,28 @@
 TableDataset = Dict[pt.TableId, pt.Dataset]
 
 
+class RuleFilter:
```
yulric, line 33: We should probably have some kind of validation here to ensure that a rule id is not in both lists? I think otherwise it could cause confusion in the future.

yulric (changes requested):
1. In an effort to start documenting functions, can you add doccomments for the `_map_cerb_errors`, `_map_aggregated_errors`, `_gen_error_entry`, `_gen_cerb_error_entry`, `_gen_aggregated_error_entry`,  and `_map_cerb_errors` since they have changed
2. Can you update the documentation for the `validate_data` function

src/odm_validation/validation.py
yulric: Sounds good.

zargot:
it should all be resolved now @yulric

PR #166, opened 2023-03-08
ignore empty values in 'unique' rule

zargot:
Empty values in 'unique' columns were considered as duplicates and caused seemingly endless lists of row numbers in the validation report.


zargot:
closing, as this is fixed in #167

PR #176, opened 2023-04-05
rewrite report error messages for consistency

### LINKED ISSUE #135: make the validation error message format consistent
It's currently very hard to read a list of different validation rule error messages from something like
`map(lambda x: x['message'], report.errors)`.

Make the first part of the messages have a consistent format:
`<table>:<row>:<column>:<rule>: <normal message>`

Example:
`sites:1:siteID:missing_mandatory_column: missing mandatory column "siteID" in table sites`

Errors should also be sorted by table/row/column/rule to make it even easier to read.

Also, quote the values, so that it's obvious what the value is if it's an empty string, etc.

zargot:
fixes #135 

Code has been written to add consistent metadata as a prefix to the error report message, and the corresponding information has been removed from the error template of each rule. Error report assets have also been updated to reflect this.

### DIFF
--- a/tests/test_reports.py
+++ b/tests/test_reports.py
@@ -63,7 +63,9 @@ class TestReports(common.OdmTestCase):
         expected_errors = [{
             'columnName': 'addID',
             'errorType': 'missing_mandatory_column',
-            'message': 'Missing mandatory column addID in table addresses',
+            'message': ('missing_mandatory_column rule violated in '
+                        'table addresses, column addID, row 2: '
+                        'Missing mandatory column addID'),
             'tableName': 'addresses',
             'validationRuleFields': []
         }]
--- a/tests/test_reports.py
+++ b/tests/test_reports.py
@@ -58,13 +58,13 @@ class TestReports(common.OdmTestCase):
             ]
         }
 
-        # just one error even tho there are 2 rows missing the
+        # just one error even tho there are 2 rows missing the column
         report = validate_data(schema, data, data_kind=DataKind.spreadsheet)
         expected_errors = [{
             'columnName': 'addID',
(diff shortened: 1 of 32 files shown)


tests/test_reports.py
```diff
@@ -63,7 +63,9 @@ def test_column_rules_reported_once_for_spreadsheet(self):
         expected_errors = [{
             'columnName': 'addID',
             'errorType': 'missing_mandatory_column',
-            'message': 'Missing mandatory column addID in table addresses',
+            'message': ('missing_mandatory_column rule violated in '
```
yulric, line 66: Shouldn't the row not be in the prefix for a spreadsheet?

yulric (changes requested):
1. For the duplicate_entries_found rule, there will always be more than 1 row in the error report. Can you update the prefix so that "row" becomes "rows"?
2. For the missing_values_found rule, we should put in the offending value in the message. Maybe change the message after the prefix to, 'Missing value "<value>" found'. To avoid confusion with empty strings, it should be, "Empty string found".

tests/test_reports.py
zargot: done

zargot:
> 1. For the duplicate_entries_found rule, there will always be more than 1 row in the error report. Can you update the prefix so that "row" becomes "rows"?
> 2. For the missing_values_found rule, we should put in the offending value in the message. Maybe change the message after the prefix to, 'Missing value "" found'. To avoid confusion with empty strings, it should be, "Empty string found".

all done

tests/test_reports.py
zargot: resolving myself, to prepare for release

zargot:
merging myself, to prepare for release

PR #177, opened 2023-04-07
fix misc issues

zargot:
prepare dev branch for release by fixing small remaining issues.


zargot:
@yulric, I'm merging this on my to get 'dev' ready for release. Let me know if there are any issues with it and we can fix it in a separate PR before releasing.

PR #182, opened 2023-04-16
docs: add spec for summary report

### LINKED ISSUE #74: spec summary report
All the data in a long table format (our main format) will have numbers as text. This means we will generate 1000s of warnings. We may want to have a "warning and errors summary" that would give a count of warnings and errors, without out the specific information about which lines the errors occurred. 

This means we would have the opinion of generating two reports: a summary report; and a detailed report. We currently follow this approach in other applications (our planning tool).

_Originally posted by @DougManuel in https://github.com/Big-Life-Lab/PHES-ODM-Validation/pull/64#pullrequestreview-1191376751_

Write specifications for the following:
- Summary report (high level)
- API function "summarize_report"
- CLI tool "summarize":
    - options: output dir or stdout, output format (yaml, markdown)
- CLI tool "validate":
    - options: output dir
    - yaml output

zargot:
fixes #74 

- add summary_report section to module_function.qmd
- add summary_report.qmd with usage example

### DIFF
--- a/assets/specs/summarize_report/schema.yml
+++ b/assets/specs/summarize_report/schema.yml
@@ -1,0 +1,10 @@
+schemaVersion: 2.0.0
+schema:
+  addresses:
+    type: list
+    schema:
+      type: dict
+      schema:
+        addID:
+          type: integer
+          coerce: integer
--- a/docs/specs/module-functions.qmd
+++ b/docs/specs/module-functions.qmd
@@ -117,6 +117,41 @@ All errors and warnings for each validation rule are documented in the [specific
     * `errors`: A list of Python dictionaries describing each error. For more information refer to the files in the [validation-rules](../validation-rules/) folder
     * `warnings`: A list of Python dictionaries describing each warning.
 
+## summarize_report
+
+Summarizes the validation report.
+
+### Arguments
+
+1. `report`: A validation report returned from `validate_data`.
+
+   - type: `ValidationReport`
+
+### Return
+
+Returns a summarized version of `report`.
+
+- type: A `ReportSummary` object with the following fields and methods:
+    * `data_version`: string consisting of the version of the ODM data
+    * `schema_version`: string consisting of the version of the validation schema used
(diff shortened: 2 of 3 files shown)


zargot:
I should probably include in the spec how valid tables say they are valid, in the summary report, unless that's something only the validation tool needs?

yulric (changes requested):
The `module-functions.qmd` file looks good but I'm looking for more from the specifications.

Looking for something similar to the specifications for the rules,
*  Text explaining why the function is needed 
* Explaining how it will work with examples that can be used as test cases.

Also, this was not in the task but I think we discussed allowing the user to summarize by tables, rows, and/or columns.

zargot:
> The `module-functions.qmd` file looks good but I'm looking for more from the specifications.
> 
> Looking for something similar to the specifications for the rules,
> 
> * Text explaining why the function is needed
> * Explaining how it will work with examples that can be used as test cases.

I made an attempt at addressing this now.

> Also, this was not in the task but I think we discussed allowing the user to summarize by tables, rows, and/or columns.

I'm not sure about how we would implement this and how useful it would be. We are currently summarizing by tables, which is fine and logical. Summarizing by rows sounds like it would undo the summarization, since aggregating rows is its main feature. I'm not sure about columns. I suppose we could prefix them with table names, so it becomes like 'sites.siteID: <rule>: <count>', otherwise I think it would be confusing. Should we postpone it?

yulric:
Also, I took a brief look at your latest commit. Can you also put in examples like in the validation rules, so I can see what the output looks like and also so I know what you will be testing. Specifically,

1. What the summary object will look like. Ignore the functions in it.
2. What the summary print out will look like.

zargot:
@yulric. Sounds good, I've made new changes.

yulric (approved):

yulric:
@zargot I pushed a commit to high level spec making things clear. Feel free to rebase things to make it all clean.

Issue #160, opened 2023-03-02
implement summary report


zargot:
depends on #74 

Yulric:
1. Think we should move all the logic for creating the summary report into a seperate function, in prepearation for the new create_summary_report function that we discussed before.
2. Looking at the new code that requires passing in the ruleId is making me think we should normalize the Error object so that it should have an error type field (Warning or Error) and a ruleId field. That way you don't have to pass it in everywhere and just get it from the object.

PR #183, opened 2023-04-18
tools/validate: add outdir option

### LINKED ISSUE #149: Update the excel validation script
* Allow the user to specify where the report files will be stored
* When there are no errors/warnings, add text to the text file saying so

zargot:
fixes #149 

- use the 'typer' package for proper param handling
- add optional outdir parameter

The second part of the task, where valid tables should say "zero errors" can be deferred to #74 & #160, which will be used in the validation tool as well.


zargot:
> Oh wow, this library is amazing! Great find!
> 
> Can we add descriptions for the function parameters so they show up when the user uses the --help option?

Thanks! Yes, I figured it out now. Let me know what you think. :)

yulric (approved):

PR #191, opened 2023-04-28
Added the specification for standardizing the validation rule API

yulric:
The main content is in `validation-rule.qmd`. Everything else are assets that the specification file uses.

### DIFF
--- a/docs/specs/validation-rule.qmd
+++ b/docs/specs/validation-rule.qmd
@@ -267,7 +267,7 @@ list in the validation report. If the `errorType` is `error` its added to
 former, and if its `warning` its added to the latter.
 
 A rule can return `None`, for example, to avoid duplicate errors in the report.
-For example, with the `greater_than_max_value` rule, the `gen_error` method 
+For example, with the `greater_than_max_value` rule, the `gen_report` method 
 could return an object only for the `max` schema key.
 
 For example, for the `greater_than_max_value` rule, this method would be called 
@@ -275,13 +275,13 @@ twice, once for the `geoLat` column value in row 1 and once for the `geoLong`
 value in row 2.
 
 It would be called with the following parameters below. Each item in the list
-has the parameters for each call of the `gen_error` method.
+has the parameters for each call of the `gen_report` method.
 
 ```{python}
 #| echo: false
 
-gen_error_parameters = import_dataset(asset("gen-error-parameters.yml"))
-pprint(gen_error_parameters)
+gen_report_parameters = import_dataset(asset("gen-error-parameters.yml"))
+pprint(gen_report_parameters)
 ```
 
 #### `transformed`
@@ -329,13 +329,13 @@ The functionality of `get_schema_keys` could be rolled into `gen_schema_values`,
 so that the latter method now returns the entire schema object, with keys
 and values.
 
-This is not possible because of the way the `validate_data` and `gen_error`
+This is not possible because of the way the `validate_data` and `gen_report`
 methods work. For both these methods the library needs to know the schema
 keys that map to each rule:
 
 1. If a rule implements its own `validate_data` method, the library needs
 to attach that method to cerberus, with the schema key associated with it.
-2. For the `gen_error` method, the library needs to know how to map each
+2. For the `gen_report` method, the library needs to know how to map each
 cerberus error object to the rule it belongs to. It uses the schema key for
 this.
 
--- a/docs/specs/validation-rule.qmd
+++ b/docs/specs/validation-rule.qmd
@@ -261,10 +261,7 @@ includes:
     5. The name of the schema key that failed validation
     6. The value of the schema key
 
-The method should return an object or `None`. Depending on the value of the
-`errorType` field in the object, its either added to the errors or warnings
-list in the validation report. If the `errorType` is `error` its added to
-former, and if its `warning` its added to the latter.
+The method can return `None` or a [`report object`](#the-report-object).
 
 A rule can return `None`, for example, to avoid duplicate errors in the report.
 For example, with the `greater_than_max_value` rule, the `gen_report` method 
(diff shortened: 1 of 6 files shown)


zargot (changes requested):
- make it clear that the validation done by rules is optional and meant for the cases where cerberus needs to be extended
- rename gen_schema to gen_schema_values, as it only returns the values
- maybe specify how the rule key is connected to cerberus. in the current impl only the first key is mapped to cerberus, unless match_all_keys is set, which causes all the keys to match the rule. match_all_keys is currently only needed for the missing_values_found rule.

review not done yet, just saving it.

docs/specs/validation-rule.qmd
```diff
@@ -0,0 +1,362 @@
...
+1. A `sites` table;
+2. A `siteID` string column, part of the `sites` table;
+3. A `geoLat` integer column, part of the `sites` table; and
+4. A `geoLong` integer column, part of the `sites` table
+
+The `greater_than_max_length` rule would generate a schema only for the
+`siteID` column in the `sites` table. The schema would include its `maxLength`
+value of **6**.
+
+#### Validating data
```
zargot, line 114: make it clear that the validation done by rules is optional and meant for the cases where cerberus needs to be extended.

docs/specs/validation-rule.qmd
```diff
@@ -0,0 +1,362 @@
...
+with the library will need to extend.
+
+By using an abstract class it becomes clear what methods need to be 
+implemented when programming a new rule. Run-time checks that show errors if 
+all methods are not implemented can be shown to the user.
+
+The `ValidationRule` class has the following methods:
+
+1. `get_schema_keys`
+2. `gen_schema`
```
zargot, line 150: rename gen_schema to gen_schema_values, as it only returns the values

docs/specs/validation-rule.qmd
```diff
@@ -0,0 +1,362 @@
...
+For example, for the `greater_than_max_length` rule:
+
+1. The `gen_schema_keys` method would return the list `['maxlength']`; and
+2. The `gen_schema` method would return the list `[6]` for the `siteID` column 
+and `None` otherwise
+
+Similiarly, if we were to implement the two methods for the 
+`greater_than_max_value` rule with the parts sheet displayed above:
+
+1. The `get_schema_keys` method would return the list `['type', 'coerce', 'max']`
```
zargot, line 198: maybe specify how the rule key is connected to cerberus. in the current impl only the first key is mapped to cerberus, unless match_all_keys is set, which causes all the keys to match the rule. match_all_keys is currently only needed for the missing_values_found rule.

docs/specs/validation-rule.qmd
```diff
@@ -0,0 +1,362 @@
...
+includes:
+    1. The name of the table the failed value is from
+    2. The index of the row the failed value is from
+    3. The name of the column the failed value is from
+    4. The row object where the failed value is from
+    5. The name of the schema key that failed validation
+    6. The value of the schema key
+
+The method should return an object or `None`. Depending on the value of the
+`errorType` field in the object, its either added to the errors or warnings
```
zargot, line 263: isn't `errorType` the name of the rule used? warnings have `warningType` instead, iirc.

docs/specs/validation-rule.qmd
```diff
@@ -0,0 +1,362 @@
...
+list in the validation report. If the `errorType` is `error` its added to
+former, and if its `warning` its added to the latter.
+
+A rule can return `None`, for example, to avoid duplicate errors in the report.
+For example, with the `greater_than_max_value` rule, the `gen_error` method 
+could return an object only for the `max` schema key.
```
zargot, line 269: i suppose this is how you're handling having multiple keys for a rule. instead of only matching the first key in the list, you're moving the responsibility to gen_report where it will ignore the wrong keys?

docs/specs/validation-rule.qmd
```diff
@@ -0,0 +1,362 @@
...
+```{python}
+#| echo: false
+
+gen_error_parameters = import_dataset(asset("gen-error-parameters.yml"))
+pprint(gen_error_parameters)
+```
+
+#### `transformed`
+
+This is an optional method used to inform the rule when a value has been
```
zargot, line 287: i'm assuming that this will only be called when the values differ?

zargot (changes requested):

docs/specs/validation-rule.qmd
yulric: The use case for this is handled by `gen_error` as mentioned in this [comment](https://github.com/Big-Life-Lab/PHES-ODM-Validation/pull/191#discussion_r1188912975).

docs/specs/validation-rule.qmd
yulric: Value or type (sorry, have to confirm this as a JS dev haha)

docs/specs/validation-rule.qmd
zargot: yes :)

docs/specs/validation-rule.qmd
zargot: :+1:

docs/specs/validation-rule.qmd
zargot: you wrote it correctly in the paragraph around line 300, so if you could correct this one it would be nice.

docs/specs/validation-rule.qmd
zargot: this is becoming part of the things we need to "think about", so I'm deferring it to later.

docs/specs/validation-rule.qmd
yulric: Should be fixed in the latest commit.

PR #192, opened 2023-05-08
add spec for impl of summarize function/cli

zargot:
adds new spec files:
- summarize-report-function.qmd
- summarize-tool.qmd
- validate-tool.qmd


docs/specs/summarize-tool.qmd
```diff
@@ -0,0 +1,97 @@
...
+
+` python tools/summarize <report-file> [options...]`
+
+### Options
+
+- `--by=[(table, column, row)],...`
+
+  A comma separated list/set of the keys to summarize by. Each key specified
+  will add a separate summarization to the summary report, but the same key
+  can't be repeated as this is a set. Specifying an empty set (with `--by=`)
```
yulric, line 18: Shouldn't specifying an empty set give an error to the user? rather than generating nothing? since the user is probably doing it by mistake. Or instead, it should summarize by table.

yulric (changes requested):
1. I don't think you addressed these parts of the spec 
    1. [Generate counts of rows/columns with errors and do the same for warnings](https://github.com/Big-Life-Lab/PHES-ODM-Validation/blob/74-summary-spec/docs/specs/summary-report.qmd#L96-L97)
    2. Sorry, this is a flip-flop. It sounded like Doug actually wanted to [generate the total errors/warnings per table, row, and column](https://github.com/Big-Life-Lab/PHES-ODM-Validation/blob/74-summary-spec/docs/specs/summary-report.qmd#L96-L97). We should add this in.
2. For the summary tools:
    2.1 Can you update the MD example with warnings and row for completeness
    2.3 For the YAML example, can you update the example to be like the one for the MD output

docs/specs/summarize-tool.qmd
zargot: The default is table, so if the user doesn't know about the parameter then it becomes summarized by table, which is fine I think?

docs/specs/summarize-tool.qmd
```diff
@@ -0,0 +1,97 @@
...
+
+```{shell}
+> python tools/summarize.py $REPORT --by=table,column --format=markdown
+
+# summary
+## errors
+### table
+sites:
+- less_than_min_value: 3
+### column
```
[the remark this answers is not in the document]
yulric: Actually, thinking about this a little more, I'm not so sure about this comment. With the amount of information (rows, column, and tables) in a dataset, it probably makes sense to not do any nesting. So maybe something like:

```
### columns

- addId
    - table: addresses
        - greater_than_max_value: 3  
```

So, when a person searches for a column/row, they know immediately which table its from without doing a lot of scrolling. What do you think?

docs/specs/summarize-tool.qmd
yulric: Yup, that make sense to me. I was talking more about if the user passes `by=`, does it still summarize by table?

zargot: if the user passes `--by=` then it's the same as calling `summarize_report(report, by={})`, so there will be no summarizations. Writing the argument like that is not something someone would do by mistake I think, since it's more logical to write something after the `=`, and if something invalid is written after the `=` then they'd get an error. Making a mistake like `--by= table` should also trigger an error, since `table` by itself isn't a valid argument.

docs/specs/summarize-tool.qmd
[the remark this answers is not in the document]
zargot: well, I think the markdown version should mirror the yaml version as much as possible, and I'm assuming that the yaml/tree version is gonna have the uppermost level on the top, so table goes first, and then everything in that table follows. I guess it will also keep it more DRY. As long as we use 4-space indents I think it's easy to scroll and see what belongs. What do you think?

docs/specs/summarize-tool.qmd
yulric: 1. By no summarizations, do you mean the summary object will be empty? Or it will just have the overview field?
2. Why is `table` by itself not a valid argument? I thought that it would generate [this](https://github.com/Big-Life-Lab/PHES-ODM-Validation/blob/74-summary-spec/docs/specs/summary-report.qmd#L84-L119) part of the spec.

yulric (approved):

PR #196, opened 2023-05-23
impl summarize_report

### LINKED ISSUE #160: implement summary report

zargot:
fixes #160 

see commits and their messages.

### DIFF
--- a/src/odm_validation/rules.py
+++ b/src/odm_validation/rules.py
@@ -31,6 +31,7 @@ from versions import Version
 
 
 RuleId = Enum('RuleId', [
+    '_all',
     '_coercion',
     'duplicate_entries_found',
     'greater_than_max_length',
--- a/src/odm_validation/summarization.py
+++ b/src/odm_validation/summarization.py
@@ -1,0 +1,282 @@
+from collections import defaultdict
+from dataclasses import dataclass
+from enum import Enum
+from itertools import groupby
+from typing import Dict, List, Set, Tuple
+# from pprint import pprint
+
+import reports
+from part_tables import TableId
+from reports import ErrorKind, ValidationReport
+from rules import RuleId
+from versions import __version__
+
+
+# types
+# -----------------------------------------------------------------------------
+
+
+Count = int
+EntityId = str
+
+
+class SummaryKey(Enum):
+    table = 1
+    column = 2
+    row = 3
+
+    def __lt__(self, other):
+        if self.__class__ is other.__class__:
+            return self.value < other.value
+        return NotImplemented
+
+
+# this is the core error/rule-to-count data structure
+ErrorCounts = Dict[RuleId, Count]
+
+# table -> entity -> error/rule -> count
+# ex: SummaryKey.column -> 'addresses' -> 'addID' -> 'invalid_type' -> 3
+EntityCounts = Dict[EntityId, ErrorCounts]
+TableCounts = Dict[TableId, EntityCounts]
+
+# simplified data structures for total counts
+TableErrorCounts = Dict[TableId, ErrorCounts]
+ErrorKindErrorCounts = Dict[ErrorKind, ErrorCounts]
+
+
+@dataclass(frozen=True)
+class Counts:
+    total_counts: ErrorCounts
+    key_counts: Dict[SummaryKey, TableCounts]
+
+
+@dataclass(frozen=True)
+class SummaryEntry:
+    rule_id: RuleId
+    count: int
+
+    key: SummaryKey
+    """The group/summary-key this entry is summarized by."""
+
+    value: str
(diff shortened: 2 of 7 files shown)


src/odm_validation/rules.py
```diff
@@ -31,6 +31,7 @@
 
 
 RuleId = Enum('RuleId', [
+    '_all',
```
yulric, line 34: Hmm, not sure this belongs here, its not a rule ID. Shouldn't it go somewhere in the summarization module?

src/odm_validation/summarization.py
```diff
@@ -0,0 +1,309 @@
...
+        'warnings': dict(errorkind_counts[ErrorKind.WARNING]),
+    }
+    return overview
+
+
+# API
+# -----------------------------------------------------------------------------
+
+
+def summarize_report(report: ValidationReport,
```
yulric, line 261: If I understand your code, your're creating an internal data structure which is then used to generate the different pieces of the summary report (overview and summary entries). Can we directly create the summary entries and create the overview from them?

The internal data structure is hard to grok with all its nested layers.

yulric (changes requested):

src/odm_validation/rules.py
zargot: well, it's "all" rule ids, so I think it fits reasonably well, kinda like a default value. If it's not part of RuleId, then I can't use RuleId as the type for 'rule_id' in the SummaryEntry class.

tests/test_summarization.py
```diff
@@ -0,0 +1,246 @@
...
+        SE(rule_id=ALL, count=1, key=SummaryKey.row, value='3'),
+    ]
+}
+
+overview = {
+    'tables': {
+        'addresses': {
+            'columns': 1,
+            'rows': 2,
+            'rules': {
```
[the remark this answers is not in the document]
zargot: hmm. Maybe I misunderstood our conversations earlier. You're talking about `overview.tables.<table>.rules`? I suppose it's the same as summarizing by table, yeah. I'll happily remove it, if you want?

src/odm_validation/summarization.py
zargot: As far as I can see, the overview is already generated from the summary entries?

tests/test_summarization.py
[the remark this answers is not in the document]
yulric: Yeah, it seems redundant to have it in there when they can generate it by passing in `by = Table`.

src/odm_validation/rules.py
yulric: Sure, I guess it doesn't really matter since all of this will probably change with the `ValidationRule` change.

src/odm_validation/summarization.py
yulric: Do we need the internal data structure? Can we not "directly" create the `ErrorSummary` object without it?

Your right about the overview.

src/odm_validation/rules.py
zargot: *my face when my code doesn't matter* xD

src/odm_validation/summarization.py
zargot: i can give it a try for sure

tests/test_summarization.py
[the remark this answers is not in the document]
zargot: done

src/odm_validation/summarization.py
zargot: i removed at least one layer of complexity. is it good now?

src/odm_validation/summarization.py
```diff
@@ -73,69 +78,80 @@ class SummarizedReport:
...
+        total_counts[rule_id] += count
+        for key in keys:
+            if key == SummaryKey.row:
+                for row_id in row_ids:
+                    entity_id = str(row_id)
+                    _update_entity(key, table_id, entity_id, rule_id, 1)
+            else:
+                column_id = e['columnName']
+                key_ix = _key_index(key)
+                entity_id = [table_id, column_id][key_ix]
```
yulric, line 138: This seems less clear then using a ternary operator.

yulric (approved):
Looks good, thanks for making those changes.

src/odm_validation/summarization.py
zargot: done :)

PR #204, opened 2023-08-16
refactor rule errors

### LINKED ISSUE #132: refactor validation report and error generation
validation.py has a lot of private functions doing the heavy lifting for the error-generation and validation-report generation. It would be better if this was moved into another module, like reports.py.

zargot:
fixes #132

### DIFF
--- a/src/odm_validation/validation.py
+++ b/src/odm_validation/validation.py
@@ -4,324 +4,30 @@ generation and data validation.
 """
 
 from collections import defaultdict
-from collections.abc import Iterable, Iterator
-from copy import deepcopy
-from itertools import groupby
-from typing import Any, Callable, Dict, List, Optional, Set, Tuple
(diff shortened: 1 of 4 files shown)


src/odm_validation/validation.py
```diff
@@ -385,6 +91,10 @@ def generate_validation_schema(parts: pt.Dataset,
 OnProgress = Callable[[str, str, int, int], None]
 
 
+def _strip_coerce_rules(cerb_schema):
+    return strip_dict_key(deepcopy(cerb_schema), 'coerce')
```
yulric, line 95: Magic string, and it looks like its hardcoded in other sport? Maybe add a task to fix this?

yulric (approved):

src/odm_validation/validation.py
zargot: #205

PR #207, opened 2023-12-01
include asset dir in pkg, and fix import-time errors

### LINKED ISSUE #206: [Bug]: failed to get latest ODM version
### Bug Description

I installed according to the instructions using pip. If I launch Python and try:
```from odm_validation.validation import validate_data``` 
I get the error "failed to get latest ODM version”

MacOS 13.6.1

### Python Version

3.10.13

### PR DESCRIPTION
**Pull request by zargot**
Attempts to fix the breaking errors occurring when importing the library.

See commits.

fixes #206

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,3 +16,11 @@ classifiers = [
 [project.urls]
 "Homepage" = "https://github.com/Big-Life-Lab/PHES-ODM-Validation"
 "Bug Tracker" = "https://github.com/Big-Life-Lab/PHES-ODM-Validation/issues"
+
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+# https://stackoverflow.com/a/75397171
+[tool.hatch.build.targets.wheel.force-include]
+"assets/validation-schemas" = "odm_validation/assets/validation-schemas"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,15 +12,21 @@ classifiers = [
     "Programming Language :: Python :: 3",
     "Operating System :: OS Independent",
 ]
+dynamic = ["dependencies"]
 
 [project.urls]
 "Homepage" = "https://github.com/Big-Life-Lab/PHES-ODM-Validation"
 "Bug Tracker" = "https://github.com/Big-Life-Lab/PHES-ODM-Validation/issues"
 
 [build-system]
-requires = ["hatchling"]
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/odm_validation/validation.py [hunk 1]
```diff
@@ -125,7 +126,7 @@ def _validate_data_ext(schema: Schema,
     assert isinstance(data, dict), (
         '`data` must be a dict. Remember to wrap the datasets in a dict with '
         'the table names as keys.')
-    assert isinstance(data_kind, DataKind), 'invalid data_kind param type'
+    assert isinstance(data_kind, Enum), 'invalid data_kind param type'
```
yulric, line 129: Not sure I understand why this was done?

pyproject.toml
```diff
@@ -12,16 +12,20 @@ classifiers = [
...
+requires = ["hatchling", "hatch-requirements-txt"]
 build-backend = "hatchling.build"
 
 # https://stackoverflow.com/a/75397171
 [tool.hatch.build.targets.wheel.force-include]
 "odm_validation" = "odm_validation"
 "assets/validation-schemas" = "odm_validation/assets/validation-schemas"
+
+[tool.hatch.metadata.hooks.requirements_txt]
+files = ["requirements.txt"]
```
yulric, line 31: Do we need to include this so that the package's dependencies get automatically installed?

pyproject.toml
```diff
@@ -16,3 +16,12 @@ classifiers = [
...
 "Homepage" = "https://github.com/Big-Life-Lab/PHES-ODM-Validation"
 "Bug Tracker" = "https://github.com/Big-Life-Lab/PHES-ODM-Validation/issues"
+
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+# https://stackoverflow.com/a/75397171
+[tool.hatch.build.targets.wheel.force-include]
+"odm_validation" = "odm_validation"
```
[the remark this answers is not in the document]
zargot: done

pyproject.toml
zargot: yeah, otherwise it won't recurse, it'll just install odm_validation but not its dependencies, so you'll end up with missing lib errors.

src/odm_validation/validation.py
zargot: yeah, the only reason for check is to make sure that the user passes in arguments in the right order, since some of them are optional, etc. It would be solved by using typing.

PR #215, opened 2023-12-07
fix report serialization

zargot:
The TableInfo class which appears in the validation report, had to be made into a TypedDict to support normal json serialization in tools using it (namely the web validation tool).

### DIFF
--- a/src/odm_validation/reports.py
+++ b/src/odm_validation/reports.py
@@ -2,6 +2,7 @@ import datetime
 from dataclasses import dataclass, field
 from enum import Enum
 from typing import Any, Dict, List, Optional, Set
+from typing_extensions import TypedDict
 # from pprint import pprint
 
 import part_tables as pt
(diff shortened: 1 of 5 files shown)


src/odm_validation/reports.py
```diff
@@ -38,18 +38,12 @@ class ErrorCtx:
...
-    rows: int
-
-
 @dataclass(frozen=True)
 class ValidationReport:
     data_version: str
     schema_version: str
     package_version: str
-    table_info: Dict[pt.TableId, TableInfo]
+    table_info: Dict[pt.TableId, dict]
```
yulric, line 46: Can you add a comment about what the shape of this dictionary should be?

zargot:
@yulric, I managed to make TypedDict work, and removed "jsons".

yulric (approved):
Looks good. I wonder if we can replace the TableInfo class with a TypedDict, seems more light weight.

PR #228, opened 2024-11-11
cli: add odm-validate cmd

zargot:


src/odm_validation/tools/validate.py
```diff
@@ -103,10 +103,25 @@ def detect_data_format(path: str) -> Optional[DataFormat]:
...
 
 
+def get_pkg_dir() -> str:
+    # aka src dir
+    tools_dir = Path(__file__).parent
+    return str(tools_dir.parent)
+
+
+def get_asset_dir() -> str:
+    '''returns pkg asset dir, falls back to dev asset dir'''
```
yulric, line 112: What is the dev assets dir? I could not find a reference to anything like that in the code.

src/odm_validation/tools/validate.py
```diff
@@ -78,7 +78,7 @@ def import_xlsx(src_file, dst_dir) -> List[str]:
 def get_sheet_table_id(schema, sheet_name) -> Optional[str]:
     table_ids = list(schema['schema'].keys())
     for table_id in table_ids:
-        if sheet_name.endswith(' ' + table_id):
+        if sheet_name.lower().endswith(table_id.lower()):
```
yulric, line 80: Did you run this by Mathew? We talked about how boolean values in the parts sheet are case-sensitive and I'm wondering if its the same with all parts including tables names.

yulric (changes requested):

src/odm_validation/tools/validate.py
zargot: oh, it's the asset dir in the repo (`/assets`). I called it the "dev dir" since it's the one we use during development, while a different dir is used after the package is installed (`odm_validation/assets`).

src/odm_validation/tools/validate.py
zargot: good point, I mainly just wanted to remove the leading space. Fixed.

src/odm_validation/tools/validate.py
yulric: Oh I see, that makes sense. Can you make it clear that its the directory used when developing the package

zargot: done

zargot:
i'll wait with merging this until the other PRs depending on this one are reviewed, since they usually go out of sync otherwise.

PR #231, opened 2024-11-11
exclude dictionary tables from validation

zargot:


zargot:
> I talked to Mathew and he said you can identify the metadata tables by filtering those that have the class column set to `lookup`. The class column was introduced in 2.0.0 so pre that version we should hardcode the list of metadata tables to be parts and sets.

great thanks! Fixed now. Mathew said that there weren't metadata tables pre v2, so I'm just doing additional checks for v2 now.

yulric (approved):

Issue #227, opened 2024-11-05
[Bug]: missing category value triggers invalid-category rule


zargot:
### Bug Description

error:
`polygons(geoType, 2): Invalid category "" [invalid_category]`

@yulric, I'm assuming this is only supposed to trigger the missing-required-value rule, if any?

### The operating system on which the bug occured

linux

### Python Version

3.8

PR #232, opened 2024-11-11
rules: ignore empty values for minlength

### LINKED ISSUE #226: [Bug]: missing phone number triggers min-length rule
### Bug Description

@yulric, shouldn't this only trigger the missing-required-value rule when it's empty, if it's required in the first place?

error:
`contacts(phone, 2): Value "" (of length 0) is less than the min length of "10" [less_than_min_length]`

linux

### Python Version

3.8

zargot:
so that the minlenght rule won't trigger unless there is a value in the first place

depends on #237 

fixes #226


zargot:
closing this in favor of a new PR that fixes both #226 and #227.

PR #235, opened 2024-11-13
fix boolean set values in schemas

### LINKED ISSUE #224: [Bug]: "TRUE" is not accepted as a boolean value
### Bug Description

parts:
image

measures:
image

error:
`measures(reportable, 2): Column reportable is a boolean but has value "TRUE". Allowed values are false/true. [invalid_type]`

linux

### Python Version

3.8

zargot:
fixes #224

### DIFF
--- a/src/odm_validation/rule_primitives.py
+++ b/src/odm_validation/rule_primitives.py
@@ -58,7 +58,6 @@ def _get_attr_meta(attr: Part, table_id: PartId, version: Version,
         if version.major >= 2:
             allowed_ids = cerb_rules.get('allowed')
             if allowed_ids:
-                # print(pt.get_partID(attr))
                 set_id = attr['mmaSet']
                 for part_id in allowed_ids:
                     result.append({'partID': part_id, 'setID': set_id})
--- a/src/odm_validation/rule_primitives.py
+++ b/src/odm_validation/rule_primitives.py
@@ -59,6 +59,11 @@ def _get_attr_meta(attr: Part, table_id: PartId, version: Version,
             allowed_ids = cerb_rules.get('allowed')
             if allowed_ids:
                 set_id = attr['mmaSet']
+
+                # XXX: bool part-ids are in lower case
+                if set_id == pt.BOOLEAN_SET:
+                    allowed_ids = list(map(str.lower, allowed_ids))
+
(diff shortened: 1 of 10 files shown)


src/odm_validation/rule_primitives.py
```diff
@@ -57,8 +57,12 @@ def _get_attr_meta(attr: Part, table_id: PartId, version: Version,
         if version.major >= 2:
             allowed_ids = cerb_rules.get('allowed')
             if allowed_ids:
-                # print(pt.get_partID(attr))
                 set_id = attr['mmaSet']
+
+                # XXX: bool part-ids are in lower case
+                if set_id == pt.BOOLEAN_SET:
```
yulric, line 63: Is this chunk adding the lower case boolean values as alloweable?

yulric (changes requested):
1. I think the dictionary for v2.0.0 should be updated to the one one [GitHub](https://github.com/Big-Life-Lab/PHES-ODM/blob/main/Archived%20V2.0/ODM_parts_v2.0.0.csv). The one you have still contains the lowercase true and false.

src/odm_validation/rule_primitives.py
zargot: `allowed_ids` is a list of part-ids, and the boolean parts have lower-case part-ids, so I have to make them lower case again here since they're retrieved from the list of allowed values which has upper-case booleans. 

This is basically an edge case I have to cover due to the ODM boolean definition having different case depending on the context.

zargot:
> 1. I think the dictionary for v2.0.0 should be updated to the one one [GitHub](https://github.com/Big-Life-Lab/PHES-ODM/blob/main/Archived%20V2.0/ODM_parts_v2.0.0.csv). The one you have still contains the lowercase true and false.

Oh, they made it consistent? That would help xD

I'm planning to update the schemas for the next version (0.7). I'll write a note on that issue to refactor this, if it's ok for you to do it this way.

src/odm_validation/rule_primitives.py
yulric: What are the different contexts which requires different versions of the boolean parts?

yulric (approved):

PR #238, opened 2024-11-19
allow empty values with minlength & allowed rules

### LINKED ISSUE #226: [Bug]: missing phone number triggers min-length rule

### LINKED ISSUE #227: [Bug]: missing category value triggers invalid-category rule

zargot:
- fixes #226
- fixes #227

see commits


src/odm_validation/rules.py
```diff
@@ -250,7 +277,10 @@ def gen_schema(data: pt.OdmData, ver: Version):
                 cat_ids1 = pt.map_ids(data.mappings, cat_ids0, ver)
                 if len(cat_ids1) == 0:
                     continue
-                cerb_rules = {cerb_rule_key: sorted(set(cat_ids1 + other_cat))}
+                cerb_rules = {'anyof': [{
+                    'empty': True,
```
yulric, line 281: Should't this be added only if the part does not allow missing values?

src/odm_validation/rules.py
```diff
@@ -201,7 +228,7 @@ def less_than_min_length():
     def gen_cerb_rules(val_ctx: OdmValueCtx):
         val = try_parse_int(val_ctx.value)
         if val > 0:
-            return {'minlength': val}
+            return {'anyof': [{'empty': True, 'minlength': val}]}
```
yulric, line 231: Shouldn't this be added only if the part does not allow missing values?

yulric (changes requested):
1. What's the reason for updating the schema and schema additions files in the module-functions folder?
2. Can you explain why you're using anyof rather than just including `empty = True` directly?

src/odm_validation/rules.py
zargot: well, I don't think we can check that since that would make rules depend on each other, but I'll make a test and see if it works, or if I can fix it.

src/odm_validation/rules.py
zargot: see other comment.

src/odm_validation/rules.py
zargot: actually, I misunderstood. `empty: true` means that empty values are allowed, so it's correct like this. I'm still making the test to see if there are any incompatibilities with the missing_values_found rule tho.

zargot: I've added the test now, see the last commit. Some warnings/errors overlap, but that's to be expected since all rules are independent.

yulric (approved):

PR #254, opened 2025-01-28
add new odm dictionary versions

zargot:


src/odm_validation/part_tables.py
```diff
@@ -165,29 +165,28 @@ def _strip_prerelease(v: Version) -> Version:
     return result
 
 
-def get_version_range(part: Part) -> tuple[Version, Version]:
-    # XXX: must have default for tests (without versioned parts) to work
-    row = part
+def get_first_released(part: Part) -> Version:
```
yulric, line 170: Can you add a doc comment for this function? Or rename it to `get_odm_version_when_first_released`?

src/odm_validation/part_tables.py
```diff
@@ -165,29 +165,28 @@ def _strip_prerelease(v: Version) -> Version:
...
+    #
+    # NOTE: pre-releases are ignored
+    # XXX: must have a default value for tests without versioned parts to work
     v = _strip_prerelease(version)
-    first, last = get_version_range(part)
-    active = is_active(part)
-    return (first <= v and v < last) or (v == last and active)
+    first = get_first_released(part)
+    latest = _strip_prerelease(odm.VERSION)
+    if is_active(part):
```
yulric, line 185: I'm assuming this function is "complicated" only because we need to take into account that version 1 parts are in the v2.0.0 schema. Otherwise the logic would be, if its active then its compatible otherwise its not?

yulric: If that's so, then can we not just check if the firstReleased is version 1 to know if its compatible with version 1?

src/odm_validation/odm.py
```diff
@@ -18,24 +16,7 @@ def get_table_names_filepath(v: Version) -> str:
...
-            continue
-        versions.append(v)
-    if len(versions) == 0:
-        sys.exit("failed to get latest ODM version")
-    versions.sort()
-    return versions
-
-
-CURRENT_VERSIONS = _get_odm_versions()
+CURRENT_VERSIONS = list(map(parse_version, ['2', '2.1', '2.2.3']))
```
yulric, line 19: Can the asset generation script not get the versions it has to parse from the assets/dictionary folder itself? Rather than using this variable?

My worry here is that we're going to forget to update this variable when there's a new version.

src/odm_validation/rule_primitives.py
```diff
@@ -66,8 +66,9 @@ def _get_attr_meta(attr: Part, table_id: PartId, version: Version,
             if allowed_ids:
                 set_id = attr['mmaSet']
 
-                # XXX: bool part-ids are in lower case
-                if set_id == pt.BOOLEAN_SET:
+                # XXX: booleanSet's part-ids are lower case in ODM v2.2 and
+                # lower
```
yulric, line 70: Seems like this is incomplete?

src/odm_validation/part_tables.py
```diff
@@ -381,16 +383,50 @@ def _table_has_attr(table: Part, attr: Part, version: Version) -> bool:
...
-    # function parts/sets validation and hotfixes
-
-    # FIXME: true/false parts are missing version1Category
-    for part_id in BOOL_PART_IDS:
-        part = all_parts.get(part_id)
-        if part and should_have_mapping(part, version, odm.VERSION):
-            if V1_CATEGORY not in part:
-                part[V1_CATEGORY] = part_id.capitalize()
-                assert has_mapping(part, version)
+def fix_parts(all_parts: PartMap, version: Version) -> None:
```
yulric, line 386: I keep forgetting, but did we decide that boolean parts are not going to be case-sensitive?

src/odm_validation/part_tables.py
```diff
@@ -381,16 +383,50 @@ def _table_has_attr(table: Part, attr: Part, version: Version) -> bool:
...
+
+    if version.major == 2:
+        if version.minor in [0, 2]:
+            # XXX: normalize bool sets' part id to upper case, since they're
+            # lower-case in ODM v2.0 and v2.2.
+            for s in sets:
+                if s[SET_ID] == BOOLEAN_SET:
+                    s[PART_ID] = s[PART_ID].upper()
+
+        if version.minor == 2 and version.patch == 3:
```
yulric, line 424: Shouldn't this be fixed in the dictionary?

yulric (changes requested):

src/odm_validation/odm.py
zargot: I wrote about it in the commit message:

>     remove dynamic load of odm versions
> 
>     It's not practical: Versions can only be detected from the dicionary
>     files, but they are not included in the package. Only files generated
>     from the dictionary are included, and those can't be used to detect the
>     versions since the asset generation script also needs to know the
>     versions (before the generated assets exist).
> 
>     A workaround could be to have two different way of detecting the
>     versions depending on the context. The generation script could read the
>     dictionaries, and the rest of the code could read the generated assets.
>     However, this is just not necessary when the hardcoded alternative is so
>     simple.

src/odm_validation/part_tables.py
zargot: they are case sensititve (UPPER only). I'm just normalizing it since different odm versions have specified it differently.

src/odm_validation/part_tables.py
zargot: sure, but it's already released, and the next version isn't out yet. I can remove this part when I overwrite v2.2.3 with v2.2.4.

src/odm_validation/rule_primitives.py
zargot: just badly worded, fixed now

src/odm_validation/part_tables.py
zargot: We're generating schemas for many different versions, and we may want to use just a single dictionary to generate those schemas. Right now we're just generating multiple versions from the v2.0 dictionary like you said (v1.0, v1.1 and v2.0), but that may change and this function covers all those cases with relatively simple logic. Without this we'd have to hardcode the edge cases, and be careful not to have multiple schema versions per dictionary or remember to add more edge cases, etc. I think the current solution is good?

src/odm_validation/part_tables.py
zargot: what about `get_initial_version`?

src/odm_validation/part_tables.py
yulric: "we may want to use just a single dictionary to generate those schemas", since we're moving towards storing each minor version of the dictionary in the repo, this point is no longer valid right (other than generating version 1 schemas from the version 2 dictionary)

"covers all those cases with relatively simple logic", the logic does not seem simple to me, which is why I brought it up. 

"Without this we'd have to hardcode the edge cases", I think its fine to do that since the only edge case is for version dictionaries which will never change.

src/odm_validation/odm.py
yulric: Oh right. I still think my point about forgetting to update the variable is valid and code is not complicated right?

src/odm_validation/part_tables.py
yulric: But in v2.0.0 and below isn't the part lower case and so shouldn't the schema also be lower case?

src/odm_validation/part_tables.py
yulric: Ok, can you add a TODO here for it

src/odm_validation/part_tables.py
zargot: well, that was an unintentional error in v2.0, afaik, so I think we should account for that? Otherwise people get an invalid validation according to the ODM spec.

src/odm_validation/odm.py
zargot: I'll take a second look.

src/odm_validation/part_tables.py
zargot: we would still need the same existing logic in the edge case for version 1, so doing this would make it even more complicated. This is because some version1 parts are active, while others are not, and they can be for v1.0.0 or v1.1.0, so it's not trivial.

src/odm_validation/odm.py
zargot: ok, I managed to get it to work. Versions are now first read from the dictionaries (to support asset generation), but when they're not present (as is the case with the package) it'll fall back to the table-name asset files.

src/odm_validation/part_tables.py
zargot: done, as discussed

yulric (approved):

PR #256, opened 2025-02-03
doc: use installed package

zargot:


yulric (commented):
sphinx still gives a warning about not recognising {bash}. It looks like you can remove the warning by changing `{bash}` to `bash`

yulric (approved):

PR #257, opened 2025-02-07
optimize summarization tool

zargot:
The reason for the immense slowness is yaml. Only json is allowed from now on.


yulric:
@zargot I'm not sure this PR is necessary, if reading a YAML file is slow then shouldn't we recommend using the JSON format rather than completely removing YAML file support?

zargot:
> @zargot I'm not sure this PR is necessary, if reading a YAML file is slow then shouldn't we recommend using the JSON format rather than completely removing YAML file support?

This PR also fixes the broken json report generation, so part of it is required in any case.

Regarding not removing yaml summarization: I feel like the slowness makes it broken and not worth using, so then I don't see the value of including it. But if you think it's worth it then I can restore it with a warning saying that it's better to use json.

yulric:
> > @zargot I'm not sure this PR is necessary, if reading a YAML file is slow then shouldn't we recommend using the JSON format rather than completely removing YAML file support?
> 
> This PR also fixes the broken json report generation, so part of it is required in any case.
> 
> Regarding not removing yaml summarization: I feel like the slowness makes it broken and not worth using, so then I don't see the value of including it. But if you think it's worth it then I can restore it with a warning saying that it's better to use json.

Lets keep the YAML feature in there, we can take a look if its too slow and users are complaining.

zargot:
fixed now, but I had to force push.

yulric (approved):

PR #260, opened 2025-02-24
fix bool value assert

zargot:


src/odm_validation/rules.py
```diff
@@ -326,7 +326,9 @@ def invalid_type() -> Rule:
     def get_error_template(odm_value: SomeValue, odm_type: Optional[str],
                            data_kind: DataKind) -> str:
         if odm_type == pt.BOOLEAN:
-            assert isinstance(odm_value, str)
+            # XXX: strings 'True' & 'False' may implicitly become booleans,
```
yulric: Can you add an example of when they can become a boolean

yulric (approved):

src/odm_validation/rules.py
zargot: the example is already there: strings 'True' and 'False' can become booleans since that's how booleans are written in python. I added that last part to the comment now. Sorry, it became a force push by accident.

zargot:
abandoning this since the issue was in the webtool, which is fixed in https://github.com/Big-Life-Lab/PHES-ODM-Validation-Web/pull/95

PR #261, opened 2025-02-26
sum: handle column errors properly

zargot:
summarization of column errors caused an exception due to missing row-numbers


yulric (approved):
Would be nice if there was a Union of all the possible errors. I assume mypy would have then figured out there are certain error objects where there aren't any row numbers.

zargot:
> Would be nice if there was a Union of all the possible errors. I assume mypy would have then figured out there are certain error objects where there aren't any row numbers.

I think it was tricky to use static typing for the errors for multiple reasons, including different error variations, different types for the same fields (like 'rowNumber' and 'rowNumbers'), and probably some serialization requirements as well. Anyway, I made an issue to investigate it. #262

PR #264, opened 2025-03-11
use ranges for error row-numbers

### LINKED ISSUE #263: convert list of numbers into ranges, in errors
this is a big issue when the list has a thousand numbers and the user is trying to read the error messages.

ex: rowNumbers: [1,2,3,4,5,6,7,8] -> rowNumbers: 1..8

zargot:
see commits.

### DIFF
--- a/assets/validation-rules/duplicate-entries-found/error-report-1.json
+++ b/assets/validation-rules/duplicate-entries-found/error-report-1.json
@@ -5,7 +5,7 @@
             "tableName": "addresses",
             "columnName": "addId",
             "invalidValue": "1",
-            "rowNumbers": [1,2],
+            "rowNumbers": [[1,2]],
             "rows": [
                 {
                     "addId": "1",
@@ -22,7 +22,7 @@
                     "addresses": "pK"
                 }
             ],
-            "message": "duplicate_entries_found rule violated in table addresses, column addId, row(s) 1,2: Duplicate entries found with primary key \"1\""
+            "message": "duplicate_entries_found rule violated in table addresses, column addId, row(s) 1..2: Duplicate entries found with primary key \"1\""
         }
     ],
     "warnings": []
--- a/src/odm_validation/reports.py
+++ b/src/odm_validation/reports.py
@@ -4,6 +4,7 @@ from enum import Enum
 from typing import Optional, TypedDict, Union, cast
 
 import odm_validation.part_tables as pt
+import odm_validation.stdext as stdext
 from odm_validation.part_tables import SomeValue
 from odm_validation.input_data import DataKind
 from odm_validation.rules import get_anyof_constraint, RuleId
@@ -12,6 +13,9 @@ from odm_validation.stdext import (
     quote,
     type_name,
 )
+
+
+IntRange = tuple[int, int]
 
 
 class ErrorKind(Enum):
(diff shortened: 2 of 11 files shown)


assets/validation-rules/duplicate-entries-found/error-report-1.json
```diff
@@ -5,7 +5,7 @@
             "tableName": "addresses",
             "columnName": "addId",
             "invalidValue": "1",
-            "rowNumbers": [1,2],
+            "rowNumbers": [[1,2]],
```
yulric, line 8: Confirming that for duplicate entries over a non-continuous range for example 1-3 and 4-6 it becomes [[1,3],[4,6]]

src/odm_validation/reports.py
```diff
@@ -236,23 +256,25 @@ def gen_rule_error(ctx: ErrorCtx,
...
     # skip row info for spreadsheet-column errors
     if ctx.data_kind == DataKind.spreadsheet and ctx.is_column:
         return error
 
     # row numbers
-    if len(ctx.row_numbers) > 1:
-        error['rowNumbers'] = ctx.row_numbers
-    else:
-        error['rowNumber'] = ctx.row_numbers[0]
+    if len(row_number_ranges) > 0:
```
yulric, line 273: Won't this add the rowNumbers field to errors like greater than max length also? Asking because only the error reports for duplicate entries found was changed

src/odm_validation/stdext.py
```diff
@@ -179,3 +179,36 @@ def keep(x: Union[dict, list], target: str) -> None:
                 swapDelete(x, i)
                 continue
             i += 1
+
+
+def gen_ranges(xs: list[int]) -> list[tuple[int, int]]:
+    '''Converts a list of numbers to a list of pairs, where each pair is the
```
yulric, line 185: can you add that a pair can contain the same numbers

tests/common.py
```diff
@@ -27,6 +27,19 @@
 ASSET_DIR = join(ROOT_DIR, 'assets')
 
 
+def fixErrorRanges(e: dict) -> dict:
+    '''fixes ranges loaded from json, which are interpreted as lists'''
```
yulric: can you specify that this is to fix tuples that are parsed as lists

yulric (changes requested):

assets/validation-rules/duplicate-entries-found/error-report-1.json
zargot: 1-3 and 4-6 is continuous (1-6). If it was 1-3 and 5-6, then it would be [[1,3],[5,6]].

src/odm_validation/reports.py
zargot: it should be the same for all error-types. I'm assuming it's a coincidence that only the duplicate-entries-found asset needed changing (since the tests are passing). The only error-report for duplicate-entries-found that I can see only uses the singular 'rowNumber', so no range-syntax was needed for it.

src/odm_validation/stdext.py
zargot: done

tests/common.py
zargot: done

yulric (approved):

PR #266, opened 2025-03-25
only use plural rows/rowNumbers in error, simplify int-ranges

zargot:
ideally this would be split into two commits, but it became too much back and forth.

reports:
- use the same json-report fix everywhere

error props:
- merge rowNumber with rowNumbers
- merge row with rows

int-ranges:
- add support for single ints, so that `1` is used instead of `(1,1)`

### DIFF
--- a/src/odm_validation/reports.py
+++ b/src/odm_validation/reports.py
@@ -1,4 +1,5 @@
 import datetime
+from copy import copy
 from dataclasses import dataclass, field
 from enum import Enum
 from typing import Optional, TypedDict, Union, cast
@@ -9,13 +10,11 @@ from odm_validation.part_tables import SomeValue
 from odm_validation.input_data import DataKind
 from odm_validation.rules import get_anyof_constraint, RuleId
 from odm_validation.stdext import (
+    IntRange,
     get_len,
     quote,
     type_name,
 )
-
-
-IntRange = tuple[int, int]
 
 
 class ErrorKind(Enum):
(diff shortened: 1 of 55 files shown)


src/odm_validation/reports.py
```diff
@@ -92,20 +116,17 @@ def join_reports(a: Optional[ValidationReport], b: ValidationReport
 
 def _range_is_one(r: IntRange) -> bool:
     '''returns true if the range has a single element'''
-    return r[0] == r[1]
-
-
-def _ranges_are_one(ranges: list[IntRange]) -> bool:
-    '''returns true if there is just a single element in the list of ranges'''
-    return len(ranges) == 1 and _range_is_one(ranges[0])
+    return isinstance(r, int) or r[0] == r[1]
```
yulric, line 119: The second if statement should not happen right? According to the PR description, `add support for single ints, so that 1 is used instead of (1,1)`

yulric (approved):

src/odm_validation/reports.py
zargot: well, `(1, 1)` would still be a valid range, I'm just adding support for writing `1` instead since it's easier.

zargot: i suppose I just meant that we will use the easy way whenever we can, but I think it's good to still support all ranges, including those with a length of one.

yulric: There isn't a different between a single int and a range where both ints are the same right?

zargot: yeah it should be treated identically.

